PR #12, opened 2021-11-29
tsh_run_output that returns out and err in a buffer

anibalportero:
In this change:
* move tsh to utils
* add new tsh_run_output function that returns out and err in a buffer


asac:
LGTM

PR #14, opened 2021-12-01
Improve usage of string.h functions to make pantavisor safer 

julianhartmer:
Improve usage of string.h functions to make pantavisor safer. Adding function snprintf_isnotrunc and sprintf_path to easily check for truncation of strings.


utils/math.h
```diff
@@ -22,6 +22,9 @@
 #ifndef PV_MATH_H
 #define PV_MATH_H
 
+#define MAX_DEC_STRING_SIZE_OF_TYPE(type) \
```
anibalportero, line 25: how about changing the name of this define to MAX_DIGIT_COUNT_OF_TYPE or similar?

config.c
```diff
@@ -332,7 +336,8 @@ static int pv_config_save_creds_to_file(struct pantavisor_config *config, char *
 	int fd;
 	char tmp_path[PATH_MAX];
 
-	sprintf(tmp_path, "%s-XXXXXX", path);
+	if (!sprintf_path(tmp_path, "%s-XXXXXX", path))
+		pv_log(WARN, "tmp_path truncated to %s", tmp_path);
```
anibalportero, line 340: maybe we could put this in a macro somewhere in utils to avoid repeating it through the code?

log.c
```diff
@@ -184,10 +181,14 @@ static void __vlog(char *module, int level, const char *fmt, va_list args)
 			int err_fd = -1;
 
 
-			snprintf(err_file, PATH_MAX, "%s/%s", log_dir, ERROR_DIR);
+			if (PATH_MAX <= snprintf(err_file, sizeof(err_file), "%s/%s",
```
anibalportero, line 184: isn't this the same we do in sprintf_path?

anibalportero (commented):
I think the change is nice but I also think all those pv_logs after sprintf_path and sprintf_isnotruc could be put in a macro somehow to avoid bloating the rest of the code. Specially if those do not alter the flow of it... Not going to oppose merging like it is though as it makes the existing code much better

config.c
julianhartmer: I thought about this. My probelm with a macro is that we cannot check whether pv_log is ready to use or not. This is the next thing I want to work on: making pv_log print to stdout until the pantavisor logger is ready so that we create the macro.

anibalportero: It's OK then :)

log.c
julianhartmer: Originally, this was an oversight on my part. But now because SNPRINTF_WTRUNC uses pv_log, I kept it this way: Better not use SNPRINTF_WTRUNC in log.c, because it could lead to infintite recursion.

PR #16, opened 2021-12-02
dm-crypt: add pv_disk with dm-crypt

parthitce:
dm-crypt: add pv_disk with dm-crypt

add initial support for dm-crypt based pv_disks. current
implementation handles only CAAM and DCP explicit.

disk defaults to persistent storage. Brief features list are,

- disks.json:   added new disks.json for handling the control over
                physical meidum to use. Refer [1] for more details.
- disk types:   There are 4 disk types introduced.
                - Directory (default): will be used to mount to pvroot as default
                - dm-crypt-versatile: use the cryptsetup utils to mount the
                  disk/file encrypted with plain text key
                - dm-crypt-dcp: specific to i.Mx2 & i.Mx6 low spec variants with
                  hardware DCP module. Compared with versatile, the key will be
                  encrypted using Hardware DCP and cryptsetup uses encrypted key blob
                - dm-crypt-caam: specific to all  modern i.Mx platforms with CAAM.
                  Encryption of the key and cryptsetup uses the black blob
- run.json:     to consume the disks to use, defined the `disk` entry
                to run.json. Refere [2] for more details.

[1]: https://docs.pantahub.com/pantavisor-disks/#pantavisor-disks
[2]: https://docs.pantahub.com/app-configuration/#disks


volumes.h
```diff
@@ -40,12 +60,15 @@ struct pv_volume {
...
+	struct pv_disk *disk;
 };
 
 void pv_volume_free(struct pv_volume *v);
 
-struct pv_volume* pv_volume_add(struct pv_state *s, char *name);
+struct pv_disk* pv_disk_add(struct pv_state *s);
+void pv_disks_empty(struct pv_state *s);
 
+struct pv_volume* pv_volume_add(struct pv_state *s, char *name, char *disk);
```
asac, line 71: we could make the diff smaller and also the function not as obscure in normal case if we add a pv_volume_add_wiith_disk() and have pv_volume_add call that function with NULL .... then we only need to use pv_volume_add_wiith_disk really where we need it...

volumes.c
```diff
@@ -92,19 +124,108 @@ void pv_volumes_empty(struct pv_state *s)
...
+
+	if (d) {
+		dl_list_init(&d->list);
+		dl_list_add_tail(&s->disks, &d->list);
+	}
+
+	return d;
+}
+
+static int pv_volumes_unmount_handler(struct pantavisor *pv, struct pv_volume *v)
```
asac, line 172: when is unmount handler called in the lifecycle? only pre shut down?
Needs implementation ...

scripts/volmount/crypt
```diff
@@ -0,0 +1,34 @@
...
+volume=$3
+loopdev=$4
+
+do_mount_disk()
+{
+	mkdir -p /data/caam
+	/lib/pv/volmount/caam-keygen import /data/caam/key importKey
+	cat /data/caam/importKey | keyctl padd logon logkey2: @u
+
+	dmsetup -v create encrypted --table "0 $(blockdev --getsz $3) crypt capi:tk(cbc(aes))-plain :36:logon:logkey2: 0 $3 0 1 sector_size:512"
```
asac, line 14: when and where is that disk formatted/created first time it is encountered?

volumes.h
parthitce: Addressed, thanks

volumes.c
parthitce: `pv_volumes_mount_handler` is the only function needed now with action type as mount/umount. Updated it now.

scripts/volmount/crypt
parthitce: Device file creation is handled in the script for 1st boot now.

parthitce:
@asac @anibalportero added for CAAM. Next step is to extend only script part to handle DCP with cryptsetup command instead of dmsetup.

Not much changes in C for that. please revisit.

loop.h
```diff
@@ -28,5 +28,7 @@
 int mount_bind(char *src, char *dest);
 int mount_loop(char *src, char *dest, char *fstype, int *loop_fd, int *file_fd);
 int unmount_loop(char *dest, int loop_fd, int file_fd);
+int get_free_loop(char *devname);
```
anibalportero, line 31: I guess we are not using these out of loop.c in the end?

parser/parser_system1.c
```diff
@@ -777,6 +865,27 @@ struct pv_state* system1_parse(struct pv_state *this, const char *buf)
 	// Parse full state json
 	ret = jsmnutil_parse_json(buf, &tokv, &tokc);
 
+	count = pv_json_get_key_count(buf, "disks.json", tokv, tokc);
+	if (count == 1) {
```
anibalportero, line 869: maybe we could return NULL and goto out if count is different than 1?

scripts/volmount/crypt
```diff
@@ -0,0 +1,143 @@
...
+	cd $dir
+	mkdir -p /data/caam
+	cp $keyname* /data/caam
+	$CAAM_KEYGEN import $keyname.bb $keyname.import
+
+	cat /data/caam/$keyname.import | /bin/keyctl padd logon logkey: @u
+
+	do_find_free_loop
+	/sbin/losetup $FREE_LOOP $dir/$img_file
+	[ $? -ne 0 ] && exit 1
```
anibalportero, line 105: I guess it would be interesting to print some error message and return different error codes like @asac is doing in https://github.com/pantavisor/pantavisor/pull/19. Then, you could use tsh_run_output to get that, which is already fixed in master I think

volumes.c
```diff
@@ -92,19 +125,100 @@ void pv_volumes_empty(struct pv_state *s)
...
+	case DISK_DIR:
+	case DISK_DM_CRYPT_VERSATILE:
+	case DISK_EXT4_UUID:
+	case DISK_EXT4:
+	case DISK_UNKNOWN:
+	default:
+		return -ENOTSUP;
+	}
+
+	tsh_run(command, 1, &wstatus);
```
anibalportero, line 208: see the above comment about tsh_run_output

scripts/volmount/crypt
```diff
@@ -0,0 +1,143 @@
...
+{
+	[ ! -f $CAAM_KEYGEN ] && exit 1
+	[ ! -f /bin/keyctl ] && exit 1
+	[ ! -f /lib/pv/volmount/caam-keygen ] && exit 1
+	[ ! -f /sbin/losetup ] && exit 1
+	[ ! -f /usr/sbin/dmsetup ] && exit 1
+	[ ! -f /bin/dd ] && exit 1
+	[ ! -f /usr/sbin/mkfs.ext4 ] && exit 1
+
+	export LD_LIBRARY_PATH=/x86_64-linux-gnu
```
parthitce, line 73: @asac keyutils libs are in here by default. I haven't adjusted the path to /usr/lib or /lib as am not sure.

loop.h
parthitce: moved the implementation to script. No changes to export it

scripts/volmount/crypt
parthitce: added debug and consumed with tsh_run_output

volumes.c
parthitce: Done, thanks

parser/parser_system1.c
parthitce: Idea is to keep it functional even when disks.json isn't added to allow the existing setup.

parser/parser_system1.c
```diff
@@ -50,6 +50,93 @@
...
+		else if(!strcmp(tmp, "dm-crypt-versatile"))
+			d->type = DISK_DM_CRYPT_VERSATILE;
+		else if(!strcmp(tmp, "dm-crypt-caam"))
+			d->type = DISK_DM_CRYPT_CAAM;
+		else if(!strcmp(tmp, "dm-crypt-dcp"))
+			d->type = DISK_DM_CRYPT_DCP;
+		else if(!strcmp(tmp, "ext4-uuid"))
+			d->type = DISK_EXT4_UUID;
+		else if(!strcmp(tmp, "ext4"))
+			d->type = DISK_EXT4;
```
asac, line 105: lets not add options we dont support... lets maybe get versatile going  ... that would be sweet to document for the rpi fanbase...

atom.mk
```diff
@@ -92,7 +92,9 @@ LOCAL_GENERATED_SRC_FILES := version.c
 LOCAL_COPY_FILES := scripts/pv_e2fsgrow:lib/pv/pv_e2fsgrow \
```
asac, line 92: In general I think such change should refer to a spec or a docs page describing the new syntax ... can we make a doc entry for the new disks approach in /docs project so it appears on docs.pantahub.com  ... which we can then refer to in this commit? we can keep that page very minimal at first, but at best it should be a bit useful from the start ....  so we can unblock the landing of course

parser/parser_system1.c
parthitce: Sure, should we also move the whole type definition handling to the script? Now it's bit bad, string -> enum -> enum again. Type can be char * IMO

anibalportero:
Looks good to me :)

atom.mk
anibalportero: In case you don't know where to put it, I'd suggest a new page in "Pantavisor reference" for disks and also add the disk key to "Container Configuration" page

parthitce: Thanks for the hint. Adding the docs and checking with local docker.

parser/parser_system1.c
parthitce: Added versatile to use with plain text key

atom.mk
parthitce: https://gitlab.com/pantacor/docs/-/merge_requests/82

atom.mk
```diff
@@ -117,5 +119,10 @@ LOCAL_LIBRARIES := cryptsetup
 
 LOCAL_COPY_FILES := scripts/volmount/dm:lib/pv/volmount/dm
 
-include $(BUILD_CUSTOM)
```
asac, line 120: you broke the other module here... keep the BUILD_CUSTOM please to end each module...

parser/parser_system1.c
```diff
@@ -50,6 +50,89 @@
...
+		if (!strcmp(tmp, "directory"))
+			d->type = DISK_DIR;
+		else if(!strcmp(tmp, "dm-crypt-versatile"))
+			d->type = DISK_DM_CRYPT_VERSATILE;
+		else if(!strcmp(tmp, "dm-crypt-caam"))
+			d->type = DISK_DM_CRYPT_CAAM;
+		else if(!strcmp(tmp, "dm-crypt-dcp"))
+			d->type = DISK_DM_CRYPT_DCP;
+		else
+			d->type = DISK_UNKNOWN;
```
asac, line 103: sholdnt we fail hard on this one too?

scripts/volmount/crypt
```diff
@@ -0,0 +1,206 @@
...
+
+	if [ $dm_type == "caam" ]; then
+		/usr/sbin/dmsetup remove $device_name
+		/bin/keyctl clear @u
+		/bin/keyctl purge @u
+		rm -rf /data
+	elif [ $dm_type == "dcp" ]; then
+		/usr/sbin/cryptsetup close $device_name
+	elif [ $dm_type == "versatile" ]; then
+		/usr/sbin/cryptsetup close $device_name
```
asac, line 101: also failing here if no match ....

parser/parser_system1.c
parthitce: hard break to out now, thanks

scripts/volmount/crypt
parthitce: else added, thanks

scripts/volmount/crypt
```diff
@@ -0,0 +1,206 @@
...
+{
+	if [ $dm_type == "caam" ]; then
+		[ ! -f $CAAM_KEYGEN ] && exit 1
+		[ ! -f /usr/sbin/dmsetup ] && exit 1
+		[ ! -f /bin/keyctl ] && exit 1
+	elif [ $dm_type == "dcp" ]; then
+		[ ! -f /lib/pv/volmount/dcp-tool ] && exit 1
+		[ ! -f /usr/sbin/cryptsetup ] && exit 1
+	elif [ $dm_type == "versatile" ]; then
+		[ ! -f /usr/sbin/cryptsetup ] && exit 1
```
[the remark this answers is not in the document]
parthitce: else added, thanks

parser/parser_system1.c
```diff
@@ -50,6 +50,91 @@
...
+			d->type = DISK_DIR;
+		else if(!strcmp(tmp, "dm-crypt-versatile"))
+			d->type = DISK_DM_CRYPT_VERSATILE;
+		else if(!strcmp(tmp, "dm-crypt-caam"))
+			d->type = DISK_DM_CRYPT_CAAM;
+		else if(!strcmp(tmp, "dm-crypt-dcp"))
+			d->type = DISK_DM_CRYPT_DCP;
+		else {
+			d->type = DISK_UNKNOWN;
+			goto out;
```
asac, line 104: shouldnt we go where we free tokv etc.?

scripts/volmount/crypt
asac: please file an issue somewhere and remember to check back once this is landed ... feels wrong to have it there indeed. most likely just atom.mk tweak needed then... later

parser/parser_system1.c
```diff
@@ -50,6 +50,91 @@
...
+
+	if (jsmnutil_parse_json(value, &tokv, &tokc) < 0) {
+		pv_log(ERROR, "wrong format filter");
+		goto out;
+	}
+
+	size = jsmnutil_array_count(value, tokv);
+	if (size <= 0) {
+		pv_log(ERROR, "empty filter");
+		goto out;
```
asac, line 67: this might leak tokv?

parser/parser_system1.c
```diff
@@ -50,6 +50,91 @@
...
+
+		if (jsmnutil_parse_json(str, &diskv, &diskc) <= 0) {
+			pv_log(ERROR, "Invalid disk entry");
+			continue;
+		}
+
+		d = pv_disk_add(s);
+		if (!d) {
+			free(diskv);
+			goto out;
```
asac, line 85: same here?

asac: anything else it might leak?

parser/parser_system1.c
parthitce:
```
out:
        if (str)
                free(str);

        if (tokv)
                free(tokv);
```
out is already freeing it. Am I missing something?

asac: not sure if str?

PR #19, opened 2021-12-05
add 'dm' volmount handler: dmverity

asac:
* add support for using volmount handler to mount LOOPIMGs.
* add dm: volmount handler in scripts/volmount/dm
* update state.c to allow volmount handler to provide a list of objects not to validate
* consolidate storage.c object validation code so all use the same function and the file path name is always available as parameter context
* add volume mount handler support in volumes.c
* struct pv_volume definition grows umount_cmd field


state.c
```diff
@@ -411,20 +414,66 @@ state_spec_t pv_state_spec(struct pv_state *s)
...
+	}
+	if (res > 0) {
+		pv_log(WARN, "command exited with error code: %d", res);
+		pv_log(DEBUG, "stdout: %s", sout);
+		pv_log(DEBUG, "stderr: %s", serr);
+		return NULL;
+	}
+
+	if (strlen(serr) > 0)
+		pv_log(WARN, "get_novalidate_list stderr output: %s", stderr);
```
anibalportero, line 443: where is this stderr variable coming from?

state.c
```diff
@@ -411,20 +414,66 @@ state_spec_t pv_state_spec(struct pv_state *s)
...
+
+	return strndup(sout, _pv_bufsize_1 - 1);;
+}
+
+#ifndef ARRAY_SIZE
+#define ARRAY_SIZE(a) (sizeof(a)/sizeof((a)[0]))
```
anibalportero, line 449: same as the one in pantavisor.h

state.c
```diff
@@ -411,20 +414,66 @@ state_spec_t pv_state_spec(struct pv_state *s)
...
 
+	char *validate_list = _pv_state_get_novalidate_list(s->rev);
+	if (validate_list)
+		pv_log(DEBUG, "no validation list is: %s", validate_list);
+
 	pv_objects_iter_begin(s, o) {
 		/* validate instance in $rev/trails/$name to match */
-		if (!pv_storage_validate_trails_object_checksum(s->rev, o->name, o->id)) {
-			pv_log(ERROR, "trails object %s with checksum %s failed", o->name, o->id);
+		char needle[PATH_MAX + 67];
```
anibalportero, line 463: why PATH_MAX + 67?

storage.c
```diff
@@ -450,16 +451,27 @@ bool pv_storage_validate_trails_object_checksum(const char *rev, const char *nam
 	int len;
 	char path[PATH_MAX];
 
+
 	len = strlen("%s/trails/%s/%s") +
 		strlen(pv_config_get_storage_mntpoint()) +
 		strlen(rev) +
 		strlen(name);
+	snprintf(path, len, "%s/trails/%s",
```
anibalportero, line 459: this path will be overwritten in line 463, right?

volumes.c
```diff
@@ -112,39 +115,55 @@ static int pv_volumes_mount_volume(struct pantavisor *pv, struct pv_volume *v)
...
+	char *handler = NULL;
+	char *name = NULL;
+	const char *partname = NULL;
+	struct stat buf;
+	int wstatus;
+	char *command;
 
 	sprintf(base, "%s/disks", pv_config_get_storage_mntpoint());
 
+	handlercut = strchr(v->name, ':');
```
anibalportero, line 129: I still think the dm: thingy would be better in json format in the state json. That way we could get rid of this code here, do the job in the parser, put it in the volume struct and take here whatever is in v

volumes.c
```diff
@@ -155,6 +174,34 @@ static int pv_volumes_mount_volume(struct pantavisor *pv, struct pv_volume *v)
 			pv_log(INFO, "mounting proper .data dir");
 			mkdir_p(mntpoint, 0755);
 			ret = mount(path, mntpoint, NULL, MS_BIND | MS_REC, NULL);
+		} else if (handler) {
```
anibalportero, line 177: I guess this would be nicer in a separate function like @parthitce does in https://github.com/pantavisor/pantavisor/pull/16

scripts/volmount/dm
```diff
@@ -0,0 +1,157 @@
...
+		;;
+	noverifylist)
+		noverifylist $2
+		;;
+	verifylist)
+		verifylist $2
+		;;
+	*)
+		echo "WRONG command line"
+		exit 1
```
anibalportero, line 155: super silly comment: we already return code 1 if there is no dm manifest

pantavisor.h
```diff
@@ -26,6 +26,10 @@
 
 #include "config.h"
 
+#ifndef ARRAY_SIZE
+#define ARRAY_SIZE(a) (sizeof(a)/sizeof((a)[0]))
```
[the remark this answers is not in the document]
asac: we have ARRAY_LEN in utils/math.h

volumes.c
asac: bu isnt that cleanup something his PR brings?

volumes.c
asac: yeah i hear you but lets try like this for now.... we will regret i  am sure :)

state.c
asac: its a path + an ID which i guess could be smaller but left it at this for now...

PR #23, opened 2021-12-11
Remove storage printf

asac:
I fixed the problems with Julians code in one commit... Lets test and land...


asac:
FYI the problem was about https://github.com/pantavisor/pantavisor/pull/23/commits/ae162cb5304ed0dd15d33c4cafb44646e615134d

PR #34, opened 2022-01-27
Refactor: create paths.h

julianhartmer:
Collect all pvr internal pathes and the mounted platform pathes in pathes.h

Update: Waiting for the pvr standalone pull request to land before merging this.


plugins/pv_lxc.c
```diff
@@ -191,22 +191,44 @@ static void pv_setup_lxc_container(struct lxc_container *c,
 		c->set_config_item(c, "lxc.log.level", log_level);
 	}
 	if (p->mgmt) {
-		c->set_config_item(c, "lxc.mount.entry", "/pv pantavisor"
-							" none bind,ro,create=dir 0 0");
-		c->set_config_item(c, "lxc.mount.entry", "/pv/logs pantavisor/logs"
-							" none bind,ro,create=dir 0 0");
-		c->set_config_item(c, "lxc.mount.entry", "/pv/user-meta pantavisor/user-meta"
-							" none bind,ro,create=dir 0 0");
+		snprintf(entry, sizeof (entry),
```
anibalportero, line 194: why not SNPRINTF_WTRUNC here?

plugins/pv_lxc.c
```diff
@@ -191,22 +191,44 @@ static void pv_setup_lxc_container(struct lxc_container *c,
...
+		c->set_config_item(c, "lxc.mount.entry", entry);
 	} else {
-		c->set_config_item(c, "lxc.mount.entry", "/pv/pv-ctrl-log pantavisor/pv-ctrl-log"
-							" none bind,rw,create=file 0 0");
-		c->set_config_item(c, "lxc.mount.entry", "/pv/pv-ctrl pantavisor/pv-ctrl"
-							" none bind,rw,create=file 0 0");
-		sprintf(entry, "/pv/logs/%s/%s pantavisor/logs none bind,ro,create=dir 0 0",
-			rev, p->name);
+		snprintf(entry, sizeof (entry),
+				"%s %s none bind,ro,create=dir 0 0",
```
anibalportero, line 213: have these bind line changed on purpose?

anibalportero: I guess not, as the platforms have to be able to write on the sockets

anibalportero (commented):
Let's try to land this first to get some stuff out of the big log MR

plugins/pv_lxc.c
julianhartmer: Let me try out SNPRINTF_WTRUNC again, there might be a problem with logs not being initialized here, so printing the warning might fail. I did the changes some time ago, so I am not sure if that was the case. I will report back soon.

plugins/pv_lxc.c
julianhartmer: The purpose in these changes here is to allow the usage of the `pathes.h` macros: Mount point destination in lxc entries are not allowed to start with `/`, so instead of mounting to `/pantavisor/pv-ctrl-log`, we have to mount to `pantavisor/pv-ctrl-log`. The changes allow us to use the defines in pathes.h, they are refactor, not functional changes.

julianhartmer: > have these bind line changed on purpose?

The purpose in these changes here is to allow the usage of the pathes.h macros: Mount point destination in lxc entries are not allowed to start with /, so instead of mounting to /pantavisor/pv-ctrl-log, we have to mount to pantavisor/pv-ctrl-log. The changes allow us to use the defines in pathes.h, they are refactor, not functional changes.

anibalportero: Sorry, didn't explained myself very well. I meant the socket bind mounts were rw but you have changed them to ro, which I think it's not correct

PR #36, opened 2022-02-04
mount cgroup2

julianhartmer:
mount cgroup2 to /sys/fs/cgroup/unified


init.c
```diff
@@ -128,6 +128,11 @@ static int early_mounts()
 	mkcgroup("pids");
 	mkcgroup("rdma");
 
+	mkdir("/sys/fs/cgroup/unified", 0555);
+	ret = mount("none", "/sys/fs/cgroup/unified", "cgroup2", 0, NULL);
+	if (ret < 0)
+		exit_error(errno, "Could not mount /sys/fs/cgroup/unified");
```
asac, line 134: we intentionally dont fail for other cgroups not being mountable and i dont think we should fail if v2 is not mountable (as that might just be a kernel without v2 and lxc in general has no problem to work without it).

while at it we should also make the errors of the other cgroup mounts log in a nicer manner (right now they spam the console with errors).

PR #39, opened 2022-02-09
Fix log.c: printing time and printing to console

julianhartmer:
Fix printing of time: printing `time(NULL)` as long without casting can crash the system. Creating helper function to get string of seconds since epos.

Fix printing to console: Use vsnprintf instead of snprintf and vprintf instead of printf in __vlog.

Qemu malta is now running again.


utils/str.c
```diff
@@ -87,3 +87,28 @@ char *pv_str_skip_prefix(char *str, const char *key)
 	}
 	return str;
 }
+
+size_t epochsecstring(char *buf, size_t len, time_t t)
+{
+	struct tm *tmp;
+	size_t ret;
+	if (!buf)
+		return ~0;
```
anibalportero, line 96: just a question, what is the ~ doing in the returns?

julianhartmer: I wanted ot return bitwise not 1 as a special value for errors. I guess I did that because I don't like the ambiguity of 0 being an error code and a return code on success. I am changing it again to 0, let's keep it in line with standard library.

anibalportero: ok thanks! rest of the change looks good to me

PR #48, opened 2022-02-18
Fix loop device starvation in crypt disk setup

asac:
Use losetup -f to find next free loop device


asac:
@anibal-portero @parthitce this fixes the crypt disk mount error on late mounts that we saw for me ...

PR #61, opened 2022-04-04
fix content length case insensitive

anibalportero:
Make our header searcher case insensitive to comply with https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2

### DIFF
--- a/ctrl.c
+++ b/ctrl.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017-2021 Pantacor Ltd.
+ * Copyright (c) 2017-2022 Pantacor Ltd.
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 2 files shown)


ctrl.c
```diff
@@ -952,12 +957,7 @@ static struct pv_cmd* pv_ctrl_read_parse_request(int req_fd)
 
 	pv_log(DEBUG, "HTTP request received: %.*s %.*s", method_len, method, path_len, path);
 
-	content_length = pv_ctrl_get_value_header_int(headers, num_headers, "Content-Length");
-	if (content_length <= 0) {
-		pv_log(WARN, "HTTP request received has empty body");
-		pv_ctrl_write_error_response(req_fd, HTTP_STATUS_BAD_REQ, "Request has empty body");
-		goto out;
-	}
```
asac, line 960: why would we not need this error path anymore?

asac (changes requested):

asac:
otherwise lgtm

ctrl.c
anibalportero: We were returning -1 from pv_ctrl_get_value_header_int when no "Content-Length" header was found. As size_t is unsigned, that was actually returning > 0, so we weren't actually getting this error for request that doesn't need "Content-Length", like GET. About not checking it for PUT and POST, I don't think it will be a problem as for those we don't use content_length in any dangerous way. We will just read 0 and exit normally if the content-length we receive is 0 or if it is not specified

PR #66, opened 2022-04-21
dm-crypt: handle mounting path with perm and rev

parthitce:
adapt dm-crypt implementation to align with using
- /<storage_path>/perm/<container>
- /<storage_path>/rev/<container>/.

Done with a help of comman path. Now dm-crypt does,
- create image file if not exist
- create device mapper if not exist
- mount to common path (currently under /media/pv/dmcrypt/<disk_name>)
- mount container paths


volumes.c
```diff
@@ -185,11 +186,28 @@ struct pv_disk* pv_disk_add(struct pv_state *s)
...
 	int ret;
 	int wstatus;
 
+	disk_name = strtok(basename(disk_path), ",");
+	if (disk_name == NULL) {
+		free(disk_path);
+		return -EINVAL;
+	}
+
+	pv_paths_storage_crypt_file(path, PATH_MAX, disk_name);
```
asac, line 203: isnt this rather a pv_paths_storage_mounted_disk_path? that gets input "handler_type" and disk_name and then is constructed like /media/pv/$handler_type/$disk_name ?

parthitce: Not sure if the current changes are fine. Please revisit @asac

parthitce:
@asac @anibal-portero couple more commits added,
- parsing failure for multiple disk entries is addressed
- verity and crypt are placed under different path

PR #70, opened 2022-04-26
Declarative kernel module loading support for BSPs

ricmm:
This adds support to a declarative interface in the BSP JSON
for kernel modules that are provided by the BSP and that expose
the said modules to the application containers depending on the
presence of a named dtb or dtbo object as passed by the bootloader.

### DIFF
--- a/ctrl.c
+++ b/ctrl.c
@@ -54,6 +54,7 @@
 #include "metadata.h"
 #include "version.h"
 #include "platforms.h"
+#include "drivers.h"
 #include "paths.h"
 #include "utils/math.h"
 #include "utils/fs.h"
@@ -74,6 +75,7 @@
 #define ENDPOINT_BUILDINFO "/buildinfo"
 #define ENDPOINT_CONDITIONS "/conditions"
 #define ENDPOINT_CONFIG "/config"
+#define ENDPOINT_DRIVERS "/drivers"
 
 #define HTTP_RES_OK "HTTP/1.1 200 OK\r\n\r\n"
 #define HTTP_RES_CONT "HTTP/1.1 100 Continue\r\n\r\n"
@@ -611,22 +613,23 @@ out:
 	return pname;
 }
 
-static bool pv_ctrl_check_sender_privileged(const char *pname)
+static struct pv_platform* pv_ctrl_get_sender_plat(const char *pname)
 {
-	bool mgmt = false;
 	struct pantavisor *pv = pv_get_instance();
 	struct pv_platform *plat;
 
 	plat = pv_state_fetch_platform(pv->state, pname);
-	if (!plat) {
+	if (!plat)
 		pv_log(WARN, "could not find platform %s in current state", pname);
-		goto out;
-	}
 
-	mgmt = plat->mgmt;
+	return plat;
+}
 
-out:
-	return mgmt;
+static bool pv_ctrl_check_sender_privileged(const char *pname)
+{
+	struct pv_platform *plat = pv_ctrl_get_sender_plat(pname);
+
+	return plat ? plat->mgmt : false;
 }
 
 static struct pv_cmd* pv_ctrl_process_endpoint_and_reply(int req_fd,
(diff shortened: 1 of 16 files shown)


ctrl.c
```diff
@@ -838,6 +841,36 @@ static struct pv_cmd* pv_ctrl_process_endpoint_and_reply(int req_fd,
...
+	} else if (pv_str_startswith(ENDPOINT_DRIVERS, strlen(ENDPOINT_DRIVERS), path)) {
+		driverkey = pv_ctrl_get_file_name(path, sizeof(ENDPOINT_DRIVERS), path_len);
+
+		if (!driverkey) {
+			pv_log(WARN, "HTTP request has bad driver alias");
+			pv_ctrl_write_error_response(req_fd, HTTP_STATUS_BAD_REQ, "Request has bad driver key name");
+			goto out;
+		}
+
+		if (!strncmp("PUT", method, method_len)) {
```
anibalportero, line 853: I don't know if I like using PUT for this one.. this kind of unload/upload ops seems more suited to a POST similar to the one we use for commands, and then put the op and driver in a json in the body like "{"op":"LOCAL_RUN","payload":"4"}". Not very strong on this one, but I'd do it like that to keep consistency.

state.c
```diff
@@ -99,6 +99,7 @@ struct pv_state* pv_state_new(const char *rev, state_spec_t spec)
 		dl_list_init(&s->jsons);
 		pv_state_init_groups(s);
 		dl_list_init(&s->conditions);
+		dl_list_init(&s->bsp.drivers);
```
anibalportero, line 102: is it necessary to clean this list in pv_state_free? A change in bsp/drivers.json will always trigger a reboot update, but, maybe for completion's sake

platforms.c
```diff
@@ -236,6 +237,19 @@ void pv_platform_add_condition(struct pv_platform *p, struct pv_condition *c)
 	}
 }
 
+void pv_platform_add_driver(struct pv_platform *p, plat_driver_t type, char *value)
+{
+	struct pv_platform_driver *d = calloc(1, sizeof(struct pv_platform_driver));
```
anibalportero, line 242: where are we freeing up this?

drivers.h
```diff
@@ -0,0 +1,80 @@
...
+	struct dl_list list;
+};
+
+const char *pv_drivers_state_str(char *match);
+int pv_drivers_state(char *match);
+int pv_drivers_load(char *match);
+int pv_drivers_unload(char *match);
+
+struct pv_driver* pv_drivers_add(struct pv_state *s, char *alias, int len, char **modules);
+void pv_drivers_remove(struct pv_driver *d);
```
anibalportero, line 54: maybe call this one from pv_state_free

platforms.h
```diff
@@ -54,6 +68,7 @@ struct pv_platform {
 	struct pv_state *state;
 	bool mgmt;
 	bool updated;
+	struct dl_list drivers;
```
anibalportero, line 71: if we hare allocating and adding drivers to this list that is separated from the one in drivers.json. Are they intended to be separate lists? Would that mean that one cannot consult platform drivers from ctrl? Maybe also that drivers are going to be loaded in state but not in platform? Maybe I'm missing something

platforms.c
asac: should be fixed now. please check @anibalportero

drivers.h
asac: i think this is done now too. check it.

ctrl.c
asac: lets keep it at PUT for now.... there is not clear winner, in REST world a PUT would about an endpoint taht updates an existin resource, while a post would be about an endpoint where you can create new documents/resources. I think module loading is like activation and deactivation that REST APIs usually do with a PUT unless its about a activate/deactivate job resource that gets POSTED and then processed.

ctrl.c
```diff
@@ -838,6 +844,67 @@ static struct pv_cmd* pv_ctrl_process_endpoint_and_reply(int req_fd,
...
+			if (!mgmt)
```
anibalportero, line 865: so in the end, only mgmt can load and unload their own drivers, right?

drivers.h
anibalportero: Looks good now. I guess we could also remove pv_drivers_remove as that's not used anywhere

PR #78, opened 2022-05-27
Bugfix/fix bad storage report

fortizc:
Fix some errors with the storage report


anibalportero:
Looks good to me :)

fortizc:
I add a couple of commits because I found some problems with the PR:
1. FIrst, two functions return and receive an `off_t` that was changed to `unsigned int`
2. the calls to `pv_log` were using `PRI64u` that was replaced by `%d`

PR #80, opened 2022-06-01
Refactor file and fs libraries

fortizc:
This is the first of three PR to complete refactor the code present on `utils/fs.h` and `utils/file.h`
This PR adds `utils/filesystem.h` with all functions to replace the current implementations.

Next step:

1. Replace library calls
2. Delete the old libraries

### DIFF
--- a/utils/filesystem.c
+++ b/utils/filesystem.c
@@ -39,26 +39,6 @@ bool pv_filesystem_path_is_directory(const char *path)
     return false;
 }
 
-static int get_directory(char *dir, const char *path)
-{
-    if (!pv_filesystem_path_exist(path)) {
-        dir = NULL;
-        return -1;
-    }
-
-    if (pv_filesystem_path_is_directory(path)) {
-        strncpy(dir, path, strnlen(path, PATH_MAX));
-        return 0;
-    } else if (errno == ENOTDIR) {
-        char copy[PATH_MAX];
-        strncpy(copy, path, strnlen(path, PATH_MAX));
-
-        char *tmp = dirname(copy);
-        strncpy(dir, tmp, strnlen(path, PATH_MAX));
-    }
-    return 0;
-}
-
 void pv_filesystem_path_sync(const char *path)
 {
     char dir[PATH_MAX] = { 0 };
--- a/utils/filesystem.c
+++ b/utils/filesystem.c
@@ -2,6 +2,7 @@
 #include <dirent.h>
 #include <errno.h>
 #include <fcntl.h>
+#include <libgen.h>
 #include <limits.h>
 #include <stdarg.h>
 #include <stdio.h>
--- a/utils/filesystem.c
+++ b/utils/filesystem.c
@@ -143,3 +143,14 @@ free_dir:
 
     return ret;
 }
+int pv_fs_path_rename(const char *src_path, const char *dst_path)
+{
+    pv_fs_path_sync(src_path);
+
+    int ret = rename(src_path, dst_path);
+    if (ret < 0)
+        return ret;
+
+    pv_fs_path_sync(dst_path);
+    return 0;
+}
--- a/utils/filesystem.c
+++ b/utils/filesystem.c
@@ -1,4 +1,5 @@
 #include "filesystem.h"
+#include "tsh.h"
 #include <dirent.h>
 #include <errno.h>
 #include <fcntl.h>
@@ -380,4 +381,19 @@ int pv_fs_file_unlock(int fd)
         ret = fcntl(fd, F_SETLK, &flock);
 
     return ret;
+}
+
+int pv_fs_file_gzip(const char *fname, const char *target_name)
+{
+	int outfile[] = { -1, -1 };
+	char cmd[PATH_MAX + 32];
+
+	snprintf(cmd, sizeof(cmd), "gzip %s", fname);
+	outfile[1] = open(target_name, O_RDWR | O_APPEND | O_CREAT);
+	if (outfile[1] >= 0) {
+		tsh_run_io(cmd, 1, NULL, NULL, outfile, NULL);
+		close_fd(&outfile[1]);
+		return 0;
+	}
+	return -1;
 }
--- a/utils/filesystem.c
+++ b/utils/filesystem.c
@@ -23,12 +23,12 @@ static void close_fd(int *fd)
     *fd = -1;
 }
 
-bool pv_fs_path_exist(const char *path)
+bool pv_filesystem_path_exist(const char *path)
 {
     return access(path, F_OK) == 0;
 }
 
-bool pv_fs_path_is_directory(const char *path)
+bool pv_filesystem_path_is_directory(const char *path)
 {
     DIR *tmp = opendir(path);
     if (tmp) {
@@ -41,12 +41,12 @@ bool pv_fs_path_is_directory(const char *path)
 
 static int get_directory(char *dir, const char *path)
 {
-    if (!pv_fs_path_exist(path)) {
+    if (!pv_filesystem_path_exist(path)) {
         dir = NULL;
         return -1;
     }
 
-    if (pv_fs_path_is_directory(path)) {
+    if (pv_filesystem_path_is_directory(path)) {
         strncpy(dir, path, strnlen(path, PATH_MAX));
         return 0;
     } else if (errno == ENOTDIR) {
@@ -59,7 +59,7 @@ static int get_directory(char *dir, const char *path)
     return 0;
 }
 
-void pv_fs_path_sync(const char *path)
+void pv_filesystem_path_sync(const char *path)
 {
     char dir[PATH_MAX] = { 0 };
 
@@ -73,12 +73,12 @@ void pv_fs_path_sync(const char *path)
     }
 }
 
-int pv_fs_mkdir_p(const char *path, mode_t mode)
+int pv_filesystem_mkdir_p(const char *path, mode_t mode)
 {
     if (!path)
         return -1;
 
-    if (pv_fs_path_exist(path)) {
+    if (pv_filesystem_path_exist(path)) {
         errno = EEXIST;
         return -1;
     }
@@ -111,7 +111,7 @@ int pv_fs_mkdir_p(const char *path, mode_t mode)
     return 0;
 }
 
-void pv_fs_path_join(char *buf, int size, ...)
+void pv_filesystem_path_join(char *buf, int size, ...)
 {
     char fmt[PATH_MAX] = { 0 };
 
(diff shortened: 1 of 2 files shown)


utils/filesystem.h
```diff
@@ -0,0 +1,125 @@
...
+/*
+ * Copy the contents of the src file descriptor to dst. If close_src is true
+ * the file pointed by src will be closed.
+ * */
+ssize_t pv_filesystem_file_copy_from_fd(int src, int dst, bool close_src);
+/*
+ * Get an extended attribute from the given file.
+ * NOTE: This function allocate a new buffer and it can be freed with free()
+ * */
+char *pv_filesystem_file_get_xattr_dup(const char *fname, const char *attr);
```
anibalportero, line 86: why pv_filesystem_file_get_xattr_dup and pv_filesystem_file_get_xattr? if we have to decide between the two, I'd prefer the second maybe?

utils/filesystem.h
```diff
@@ -0,0 +1,125 @@
...
+bool pv_filesystem_path_is_directory(const char *path);
+/*
+ * Sync the given path. If path points to a file,
+ * then the parent folder is synced
+ * */
+void pv_filesystem_path_sync(const char *path);
+/*
+ * Path concatenation, the value is returned on the given buffer
+ * */
+void pv_filesystem_path_join(char *buf, int size, ...);
```
anibalportero, line 45: maybe path_concat is better?

utils/filesystem.c
```diff
@@ -0,0 +1,414 @@
...
+    vsnprintf(buf, PATH_MAX, fmt, list);
+    buf[strnlen(buf, PATH_MAX) - 1] = '\0';
+
+    va_end(list);
+}
+
+int pv_filesystem_path_remove(const char *path, bool recursive)
+{
+    if (!recursive) {
+        return remove(path);
```
anibalportero, line 136: don't forget about sync after remove here and in the next one

utils/filesystem.h
```diff
@@ -0,0 +1,125 @@
...
+ */
+
+#ifndef UTILS_FILESYSTEM_H
+#define UTILS_FILESYSTEM_H
+
+#include <stdbool.h>
+#include <sys/types.h>
+
+/*
+ * Check if the given path exist
```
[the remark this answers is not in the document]
fortizc: Ok, no problem, so I check it and remove all the redundant or not necessary comments

utils/filesystem.c
```diff
@@ -0,0 +1,414 @@
...
+    if (fd > 0)
+        close_fd(&fd);
+
+    ret = pv_filesystem_path_remove(tmp, false);
+    pv_filesystem_path_sync(tmp);
+
+    return ret;
+}
+
+ssize_t pv_filesystem_file_copy_from_fd(int src, int dst, bool close_src)
```
anibalportero, line 220: is to and from, so I'd keep just copy_fd

utils/filesystem.c
```diff
@@ -0,0 +1,414 @@
...
+    char buf[4096] = { 0 };
+    ssize_t read_bytes = 0;
+    ssize_t write_bytes = 0;
+
+    while (read_bytes = read(src, buf, 4096), read_bytes > 0)
+        write_bytes += write(dst, buf, read_bytes);
+
+    if (close_src)
+        close_fd(&src);
+
```
anibalportero, line 234: we have to sync too. In general, we sync because we can have some problems with files that should or shouldn't exist in low spec devices with slow disk throughput, so check if we need more in the rest of the code

utils/filesystem.c
```diff
@@ -0,0 +1,414 @@
...
+    return ret;
+}
+
+int pv_filesystem_file_unlock(int fd)
+{
+    struct flock flock;
+
+    flock.l_whence = SEEK_SET;
+    /*Lock the whole file*/
```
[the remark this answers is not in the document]
fortizc: Yeah... I like the `//` too, this function has been copy-pasted from the old code so I think that the /* */ style was "the rule" it would be a pleasure to change it!

utils/filesystem.c
```diff
@@ -0,0 +1,414 @@
+#include "filesystem.h"
```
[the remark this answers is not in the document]
fortizc: Yep, I plan to make the switch to `fs` when the current `fs` library is removed

utils/filesystem.c
```diff
@@ -0,0 +1,414 @@
...
+    close(*fd);
+    *fd = -1;
+}
+
+bool pv_filesystem_path_exist(const char *path)
+{
+    return access(path, F_OK) == 0;
+}
+
+bool pv_filesystem_path_is_directory(const char *path)
```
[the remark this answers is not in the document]
fortizc: Ok, this is a problem I had when I thought about this function and `pv_filesystem_path_exist`. The thing is, `pv_filesystempath_is_directory` checks if the current path is a directory, so return `false` if the path refers to a file. If I change the name to `pv_filesystem_dir_exist`, you could try to use the function to check if it's necessary to create a folder, something like this:
```C
if (!pv_filesystem_dir_exist(path))
    pv_filesystem_mkdir_p(path);
```
So, if `path` is a file, the code has a bug because we cannot create a folder and file with the same name.
That's the reason to create these two functions

utils/filesystem.c
```diff
@@ -0,0 +1,414 @@
...
+            if (mkdir(cur_path, mode) != 0 && errno != EEXIST)
+                return -1;
+            created = i + 1;
+        }
+        ++i;
+    }
+
+    // create the last directory, this happend when the given path
+    // doesn't finish with '/'
+    if (created < strlen(path)) {
```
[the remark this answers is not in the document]
fortizc: Yes, you're right it will be changed

utils/filesystem.c
fortizc: Ok

utils/filesystem.c
fortizc: Ok, I will kept an eye on it

utils/filesystem.c
```diff
@@ -0,0 +1,414 @@
...
+    return cur_size == size ? 0 : -1;
+}
+
+char *pv_filesystem_file_get_xattr_dup(const char *fname, const char *attr)
+{
+    ssize_t size = getxattr(fname, attr, NULL, 0);
+    if (size < 1)
+        return NULL;
+
+    char *val = malloc(size);
```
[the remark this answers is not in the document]
fortizc: Yes right, reading the documentation of `getxattr()` does not specify that the value be a null-terminated string

utils/filesystem.h
fortizc: `concat` is fine to me, I will change it

utils/filesystem.h
fortizc: Yeah, the difference is one receives a buffer and the other allocates a buffer for you.
In the old library (file.h), this function receives a pointer to function to allocate memory on the value field but is it always called with NULL and a previously allocated buffer. The problem with that is `getxattr` needs the size of the attribute value. The documentation says that you can call `getxattr` with `size = 0` to retrieve the size but if the buffer was allocated before the size can be smaller than it's needed.
This is the code in `file.h`

```c
int pv_file_get_file_xattr(const char *filename, char *attr, char **dst, int(*alloc)(char**, int))
{
  int val_len = -1;

  val_len = getxattr(filename, attr, NULL, 0);
  if (val_len > 0) {
    int ret = 0;

    // if you use a NULL pointer this never happen
    if (alloc)
      ret = alloc(dst, val_len + 1);
    // even with a NULL pointer ret = 0 
    if (!ret)
      val_len = getxattr(filename, attr, *dst, val_len);  // var_len was come from the previous getxattr call
  }                                                                              // so that value is not related with dst.  
  return val_len < 0 ? -errno : val_len;
}

```
Maybe, as you said, the second option is the best. If we are OK, I can remove the first one

utils/filesystem.c
```diff
@@ -0,0 +1,414 @@
...
+{
+    if (!path)
+        return -1;
+
+    if (pv_filesystem_path_exist(path)) {
+        errno = EEXIST;
+        return -1;
+    }
+
+    if (errno != 0 && errno != ENOENT)
```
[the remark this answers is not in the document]
fortizc: Fixed in 827bc0c0dddcc6df9ad4405438ba3f7a37999f7c

utils/filesystem.c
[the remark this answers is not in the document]
fortizc: Fixed in 26cd4850fbb82bc64ef4d9391d2aab464aef80ed

utils/filesystem.c
fortizc: Fixed in 9f6a26a6120497cdddadee5bfb66f381b8a9b5e6

utils/filesystem.c
fortizc: Renamed in 75c012cc78865dcc08f62f050dbf5f0db2b43572

utils/filesystem.h
anibalportero: yeah, let's do that. Then we can see how to fit those in the current code in the next PR

utils/filesystem.c
```diff
@@ -0,0 +1,414 @@
...
+        strncpy(dir, tmp, strnlen(path, PATH_MAX));
+    }
+    return 0;
+}
+
+void pv_filesystem_path_sync(const char *path)
+{
+    char dir[PATH_MAX] = { 0 };
+
+    if (get_directory(dir, path) != 0)
```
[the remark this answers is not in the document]
anibalportero: hmm, ok, let's better use the stack in this case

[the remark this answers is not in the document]
fortizc: Fixed in 63262a06299f048e0e815bfb1fef60012f39ab7b

utils/filesystem.c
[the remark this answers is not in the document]
fortizc: Fied in e7aab3009a251e62a8f9a4eae6a162729ecd19b1

utils/filesystem.h
fortizc: Removed in 1616ea96eca28383bb3904e4f009bdba3ad23d75

utils/filesystem.h
fortizc: Rename it 886df2e9a5b50522bfcf3c5b5d3984f78cc4a15d

utils/filesystem.c
fortizc: Here cannot sync, because we haven't a path, so maybe it's a good idea only to warn about this issue.
In `/proc/...` but it feels a bit hacky and not very failure proof

utils/filesystem.h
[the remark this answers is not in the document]
fortizc: Done in b258c60209244ecd60e6adf270ca379a4288237c

PR #87, opened 2022-06-16
Replace FS library

fortizc:
This PR replaces in the code the old `fs.h` and `file.h` with the new `fs.h` library


asac:
I think this is close. please make a single commit rebased on master out of it and while doing that go through all diff lines and see which are actually wanted or which just happen "accidentally".

PR #90, opened 2022-06-20
Fix warning comparing signed and unsigned types

fortizc:


anibalportero:
should we close this one then?

PR #96, opened 2022-06-25
Fix "empty filter" error

fortizc:
Add function to check if a json array is empty

Fix error checking driver list

### DIFF
--- a/parser/parser_system1.c
+++ b/parser/parser_system1.c
@@ -605,21 +605,21 @@ static int parse_platform_drivers(struct pv_state *s, struct pv_platform *p, cha
 	}
 
 	value = pv_json_get_value(buf, "required", tokv, tokc);
-	if (value) {
+	if (!pv_json_array_empty(value)) {
 		platform_drivers_add(p, DRIVER_REQUIRED, value);
 		free(value);
 		value = 0;
 	}
 
 	value = pv_json_get_value(buf, "optional", tokv, tokc);
-	if (value) {
+	if (!pv_json_array_empty(value)) {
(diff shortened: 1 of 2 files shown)


parser/parser_system1.c
```diff
@@ -605,21 +605,21 @@ static int parse_platform_drivers(struct pv_state *s, struct pv_platform *p, cha
 	}
 
 	value = pv_json_get_value(buf, "required", tokv, tokc);
-	if (value) {
+	if (!pv_json_array_empty(value)) {
```
anibalportero, line 608: I don't think this is the best solution. Imagine value is "[]". Then the buffer would not be freed! In my opinion, checking if the array is empty is already done when we do the jsmnutil_array_count later in platform_drivers_add. It's just that it should be a WARN instead of ERROR, or maybe we can even remove the error if we get 0 elements in the array. @asac might have a better insight as he implemented this :)

fortizc: As those 3 blocks in this function are pretty similar, what if we create a function to replace them and pass as a parameter the type of the drivers (required, optional or manual) and in the end, we can use something like
```C
if (value)
    free(value);
```
Maybe to avoid passing 2 parameters (a string with the name and the enum) we can implement a enum-to-string function in `platforms.h`

anibalportero: What I mean is pv_json_array_empty parses buf, which is already in code in platform_drivers_add. I propose we either remove the empty filter message or set it to WARN

asac:
do we want to do something here still?

PR #98, opened 2022-06-28
fix: usermeta does not unescape all special characters 

fortizc:
Some change was added to this PR:
- Add better function to unescape chars and replace the old `pv_str_unescape_to_ascii` from `utils/str.h`
- Change the strategy of how user-meta is unescaped, now the data is unescaped after parsing

Both changes help to fix a nasty bug I found if you create a variable with any name but with the value `ooo\nooo` (something with a `\n` in between) the device will freeze completely.

### DIFF
--- a/metadata.c
+++ b/metadata.c
@@ -462,6 +462,7 @@ static int pv_usermeta_parse(struct pantavisor *pv, char *buf)
 			break;
 
 		strncpy(value, um+(*key_i+1)->start, n);
+		pv_str_unescape_to_ascii(value, n);
 
 		// add or update metadata
 		// primitives with value 'null' have value NULL
@@ -864,17 +865,13 @@ int pv_metadata_factory_meta(struct pantavisor *pv)
 void pv_metadata_parse_usermeta(char *buf)
 {
 	struct pantavisor *pv = pv_get_instance();
-	char *body, *esc;
+	char *body = strdup(buf);
 
-	body = strdup(buf);
-	esc = pv_str_unescape_to_ascii(body, "\\n", '\n');
-	pv_usermeta_parse(pv, esc);
+	pv_usermeta_parse(pv, body);
 
 	if (body)
 		free(body);
-	if (esc)
-		free(esc);
-	// clear old
+
 	usermeta_clear(pv);
 }
 
--- a/utils/str.c
+++ b/utils/str.c
@@ -39,40 +39,41 @@ char *pv_str_replace_char(char *str, int len, char which, char what)
 	return str;
 }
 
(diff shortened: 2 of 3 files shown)


metadata.c
```diff
@@ -422,6 +422,29 @@ int pv_metadata_rm_usermeta(const char *key)
...
+	int i = -1;
+	int j = 0;
+
+	while (i < size) {
+next:
+		++i;
+		if (buf[i] == '\\') {
+			for (size_t k = 0; k < sizeof(chars); ++k) {
+				if (buf[i + 1] == chars[k])
+					goto next;
```
anibalportero, line 437: better use continue here, right?

metadata.c
```diff
@@ -422,6 +422,29 @@ int pv_metadata_rm_usermeta(const char *key)
...
+{
+	const char chars[] = {'\"', '\\'};
+	int i = -1;
+	int j = 0;
+
+	while (i < size) {
+next:
+		++i;
+		if (buf[i] == '\\') {
+			for (size_t k = 0; k < sizeof(chars); ++k) {
```
anibalportero, line 435: this is confusing to me, what's the relationship between the chars[] index and sizeof(chars)?

metadata.c
```diff
@@ -422,6 +422,29 @@ int pv_metadata_rm_usermeta(const char *key)
 	return -1;
 }
 
+static void pv_unescape_data(char *buf, int size)
```
anibalportero, line 425: I find this code a bit weird in general. Also, should it be somewhere in utils/ instead of here?

anibalportero (commented):
Sorry, out of the loop. What are we trying to achieve here?

asac:
FYI, the problem that this is trying to solve is that if i put in a VAR="blah" in the user-meta, the " will be escaped on disk in user-meta folder.... e.g. VAR=\"blah\" is what will be in the file.

metadata.c
fortizc: As `sizeof` returns the size in bytes so char is a 1-byte data type the `sizeof` is the number of elements.
Now is true, `chars` maybe isn't the best name for that array

metadata.c
fortizc: Noy really, because I need to break the inner loop and jump to the beginning of the outer loop, So the alternative is mixing `break` with a flag to avoid the execution of the last part of the outer loop

metadata.c
anibalportero: str.h I'd say :)

anibalportero:
looks good to me now. Please, test it on device

PR #100, opened 2022-07-04
set hostname to device-id

asac:


asac:
This is a draft and shouldnt land. I would like to get something more sensible than (none) as hostname, but that still allows to set hostname through mgmt apps or maybe through cmdline/config? Lets discuss. Also what name to guess in local XP?

storage.c
```diff
@@ -1058,11 +1058,13 @@ static int pv_storage_init(struct pv_init *this)
 		pv->unclaimed = true;
 		if (pv_fs_file_save(path, "", 0444) < 0)
 			pv_log(WARN, "could not save file %s: %s", path, strerror(errno));
+		sethostname("UNNAMED", strlen("UNNAMED"));
 	} else {
 		pv->unclaimed = false;
 		SNPRINTF_WTRUNC(tmp, sizeof (tmp), "%s\n", pv_config_get_creds_id());
 		if (pv_fs_file_save(path, tmp, 0444) < 0)
 			pv_log(WARN, "could not save file %s: %s", path, strerror(errno));
+		sethostname(tmp, strlen(tmp);
```
fortizc, line 1067: Missing close parenthesis?

asac: yep

PR #105, opened 2022-07-08
add CMD_LOCAL_APPLY which stages an update for next reboot without do…

asac:
…ing the actual reboot. To improve experience we allow to replace updates that are in state UPDATE_APPLIED to allow changing pointers


ctrl.h
```diff
@@ -32,8 +32,9 @@ typedef enum {
 	CMD_POWEROFF_DEVICE = 3,
 	CMD_TRY_ONCE = 4,
 	CMD_LOCAL_RUN = 5,
-	CMD_MAKE_FACTORY = 6,
-	CMD_RUN_GC = 7,
+	CMD_LOCAL_APPLY = 6,
```
anibalportero, line 35: we have to add the cmd string to pv_ctrl_string_cmd_operation in ctrl.h to make this work. Also, you can check pv_ctrl_check_command in ctrl.c to return an error if an update is going on when issuing the apply command

PR #115, opened 2022-07-28
Add sysinfo fields to devmeta

fortizc:
See the description of the fields here https://man7.org/linux/man-pages/man2/sysinfo.2.html


asac:
merged manually locally...

PR #118, opened 2022-07-29
support shipping .json files in state json as objects

asac:
We skip expanding those .json files then accordingly.

Introduce a helper function for pv_is_sha256_hex_string

### DIFF
--- a/storage.c
+++ b/storage.c
@@ -812,6 +812,15 @@ int pv_storage_meta_expand_jsons(struct pantavisor *pv, struct pv_state *s)
 		value[n] = 0;
 		snprintf(value, n, "%s", buf + (*k + 1)->start);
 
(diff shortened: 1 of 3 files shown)


storage.c
```diff
@@ -812,6 +812,15 @@ int pv_storage_meta_expand_jsons(struct pantavisor *pv, struct pv_state *s)
 		value[n] = 0;
 		snprintf(value, n, "%s", buf + (*k + 1)->start);
 
+		// also skip unpacking this if json is a sha256 hex encoded string
+		if ((*k + 1)->type == JSMN_STRING &&
+		    pv_is_sha256_hex_string(value)) {
+			free(key);
```
anibalportero, line 818: not related to this PR but, are we freeing key and value if we complete the iteration of the loop? memory leak?

PR #121, opened 2022-08-03
remember uploaded objects on first boot object sync retries

asac:


anibalportero:
I guess this will save us some POSTs to /objects. CI is failing because of clang, remember that you have to use clang-format-11 and other versions could yield in different results

PR #123, opened 2022-08-08
add configurable restart_policy at container level in run.json

anibalportero:
This PR adds restart_policy to containers in run.json, which can be set to 'system' or 'container' to perform a full board reboot or a individual container reset. This value will override the previous behavior if any object or json is modified during an update, but it preserves the old behavior based on groups when not explicitly defined.

List of changes:
* parser: add restart_policy to run.json with 'system' and 'container' options
* platform: struct now stores restart_policy
* platform: json for pvcontrol now shows restart_policy value
* state: set 'system' default restart_policy if not explicitly configured for data, root and platform containers
* state: set 'container' default restart_policy if not explicitly configured for app containers
* state: decide whether to reboot or not for an update transition depending on container restart_policy

### DIFF
--- a/state.c
+++ b/state.c
@@ -470,6 +470,51 @@ static void pv_state_set_default_conditions(struct pv_state *s)
 	pv_state_set_default_conditions_group(s, a, NULL, "STARTED");
 }
 
+static bool pv_state_group_requires_reboot(struct pv_group *g)
+{
+	int len;
+	char *name;
+
+	if (!g || !g->name)
+		return true;
+
+	name = g->name;
+	len = strlen(name);
+
+	if (pv_str_matches(name, len, "data", strlen("data")) ||
+	    pv_str_matches(name, len, "root", strlen("root")) ||
+	    pv_str_matches(name, len, "platform", strlen("platform"))) {
+		return true;
+	}
+
+	return false;
+}
+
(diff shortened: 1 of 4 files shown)


state.c
```diff
@@ -470,6 +470,52 @@ static void pv_state_set_default_conditions(struct pv_state *s)
...
+	if (!g || !g->name)
+		return true;
+
+	name = g->name;
+	len = strlen(name);
+
+	// some groups need reboot to keep backwards compatibility with runlevel behavior
+	if (pv_str_matches(name, len, "data", strlen("data")) ||
+	    pv_str_matches(name, len, "root", strlen("root")) ||
+	    pv_str_matches(name, len, "platform", strlen("platform"))) {
```
asac, line 487: imo this legacy behaviour should only apply if no groups.json has been loaded/defined. This brings us closer to the spec where our default behaviour is actually realized through an implicit groups.json

anibalportero: it will be like that, but groups.json will not be added until the third drop of this series. For now, explicit restart_policy in run.json overrides whatever group the container belongs to, but we keep backwards compatibility and the policy will be set according to the default groups ('system' to data, root and platform; 'container' to app) if not defined

PR #125, opened 2022-08-10
shutdown pantavisor in case of ctrl-c in appengine mode

anibalportero:


init.c
```diff
@@ -484,6 +496,7 @@ int main(int argc, char *argv[])
 	if (getpid() != 1) {
 		// we are going to use this thread for pv
 		pv_pid = getpid();
+		signal(SIGINT, shell_handler);
```
asac, line 499: lets adjust the comment abovce so its clear that this is also implied to be APPENGINE mode (not just executed from shell)....

anibalportero: ok

PR #126, opened 2022-08-12
make writable and exports configurable; move the rest of cmounts out appengine init

anibalportero:


init.c
```diff
@@ -408,6 +422,10 @@ static void parse_options(int argc, char *argv[], char **config_path,
 
 #define SIZE_CMDLINE_BUF 1024
 
+// somehow this should for appengine really take the
+// argv of the process instead of the /proc/cmdline
+// in turn appengine should make a file cmdline and mount
+// it into the containers as containers use that often
```
asac, line 428: this was a my TODO comment? wnate to keep it in? if so we should use XXX: to mark it as a todo or remove it again...

init.c
```diff
@@ -490,7 +509,7 @@ int main(int argc, char *argv[])
 		goto loop;
 	}
 
-	mount_cgroups();
+	other_mounts();
```
asac, line 512: i dont like the name other_mounts.... feels like we havent thought enough what these mounts actually are :)

init.c
anibalportero: it was your TODO comment yeah. Removing it as we already support this using the --cmdline arg on the init binary, or using the -c arg on the appengine run script

init.c
anibalportero: not sure about how to rename it. Maybe we can divide the function and call them mount_disks (for /writable and /exports) and mount_home (/root and /run?)

PR #127, opened 2022-08-13
refactor shutdown to be faster and add umounts

anibalportero:
Add umounts to shutdown to free up the paths we are using after execution in appengine. The shutdown function has changed too to issue a lenient kill to the platform and ph logger forks first and then check if they are still running and finally send a force kill.

List of changes:
* init: new function to umount /writable and /exports
* logs: new function to umount logs in /pv
* logserver: modiy stop function to check if pid is still running after a timeout and issue a force kill if necessary
* mount: new function to umount devemta, usermeta and dropbear in /pv
* pantavisor: shutdown now issues lenient stop for ph logger and platforms at the same time. Then, check and force are sent if needed
* pantavisor: shutdown now umounts everything beyond /storage
* pantavisor: shutdown does not concern about umounting, socket closing and freeing up memory if a reboot or power off is ordered
* ph_logger: divide stop function in lenient and force
* state: divide stop function in lenient and force
* storage: new function to umount /storage
* utils/system: divide kill pid function in lenient and force

### DIFF
--- a/pantavisor.c
+++ b/pantavisor.c
@@ -59,6 +59,7 @@
 #include "paths.h"
 #include "ph_logger.h"
 #include "logserver.h"
+#include "mount.h"
 #include "parser/parser.h"
 #include "utils/timer.h"
 #include "utils/fs.h"
@@ -222,9 +223,8 @@ static pv_state_t _pv_run(struct pantavisor *pv)
 		pv_log(INFO, "transitioning...");
 
 		pv_logserver_stop();
-		ph_logger_stop(pv);
-
-		pv_log_start(pv, pv->update->pending->rev);
+		ph_logger_stop_lenient();
+		ph_logger_stop_force();
 
 		pv_state_transition(pv->update->pending, pv->state);
 	} else {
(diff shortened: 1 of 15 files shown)


pantavisor.c
```diff
@@ -813,44 +813,47 @@ static void pv_remove(struct pantavisor *pv)
...
+	pv_log(INFO, "preparing '%s'...", shutdown_type_string(t));
+	if (pv_config_get_system_init_mode() == IM_APPENGINE)
+		pv_log(INFO,
+		       "will not actually perform '%s' as we are in appengine mode",
 		       shutdown_type_string(t));
 
+	wait_shell();
+
 	if (REBOOT == t)
 		pv_wdt_start(pv);
```
asac, line 828: this shouldnt happen in appengine either imo.

pantavisor.c
```diff
@@ -813,44 +813,47 @@ static void pv_remove(struct pantavisor *pv)
...
-		       "closing application because of appengine init mode...");
-		pv_bootloader_remove();
-		pv_remove(pv);
-		goto out;
-	}
+	// at this point, we can shutdown if not in appengine
+	if (pv_config_get_system_init_mode() != IM_APPENGINE)
+		reboot(shutdown_type_reboot_cmd(t));
 
-	reboot(shutdown_type_reboot_cmd(t));
```
asac, line 851: is reboot never returning ... or do we wnat to run through the below functions in all cases?

pantavisor.c
anibalportero: it will execute but won't start as we have disabled wdt from pantavisor.config in appengine. I'll change it here too though so we can prevent problems

pantavisor.c
anibalportero: it's one of the things I was doubtful about. The idea is, in case of actual reboot or poweroff, we rely on those so we don't need to umount, close fds or freeup memory, which is only actually needed in case of appengine, as it is not triggering reboots or poweroffs for now

PR #134, opened 2022-08-31
Feature/lxc console log

fortizc:
* Adds to the logserver the ability to subscribe/unsubscribe a file descriptor for:
  - Add data to the pv.log (when single file mode is set)
  - Create an isolated file with the data in a directory named like the "source" of the fd (if the source is a/b/c the folders will be created).
* Add API in the pv_lxc plugin to get the console log fd from lxc.
* Add API to set a pipe to capture lxc logs (in the start function).

### DIFF
--- a/logserver.c
+++ b/logserver.c
@@ -38,6 +38,8 @@
 #include <stdarg.h>
 #include <fcntl.h>
 #include <time.h>
+#include <inttypes.h>
+#include <glob.h>
 
 #include "utils/fs.h"
 #include "utils/fs.h"
@@ -45,6 +47,7 @@
 #include "utils/system.h"
 #include "pvctl_utils.h"
 #include "bootloader.h"
+#include "config.h"
 
 #ifdef DEBUG
 #define WARN_ONCE(msg, args...)                                                \
@@ -70,6 +73,8 @@
 #define PH_LOGGER_MAX_EPOLL_FD (50)
 #define LOGSERVER_FLAG_STOP (1 << 0)
 #define LOGSERVER_BACKLOG (20)
+#define LOGSERVER_MAX_EV (5)
+#define LOGSERVER_MAX_HEADER_LEN (50)
 
 #define LOG_PROTOCOL_LEGACY 0
 
@@ -86,12 +91,37 @@ struct logserver_msg {
 	char buffer[0];
 };
 
+struct logserver_fd {
+	char *platform;
+	char *src;
+	int fd;
+	struct dl_list list;
+};
+
+struct logserver_output {
+	bool sfile;
+	bool ftree;
+};
+
+struct logserver_log {
+	int maxsize;
+	int maxfile;
+};
+
 struct logserver {
-	pid_t service_pid;
+	pid_t pid;
 	int flags;
-	int epoll_fd;
-	int sock_fd;
+	int epfd;
+	int logsock;
+	int fdsock;
 	char *revision;
+	struct logserver_log log;
+	struct logserver_output out;
+	// logserver_fd
+	struct dl_list fdlst;
+	// tmp store for fd returned by connect
+	// only if was sent to the fd_sock
+	struct dl_list tmplst;
 };
 
 struct logserver_msg_data {
@@ -106,16 +136,72 @@ struct logserver_msg_data {
 	char *data;
 };
 
-static struct logserver logserver_g = { .service_pid = -1,
+static struct logserver logserver_g = { .pid = -1,
 					.flags = 0,
-					.epoll_fd = -1,
-					.sock_fd = -1,
+					.epfd = -1,
+					.logsock = -1,
+					.fdsock = -1,
+					.log = { .maxsize = -1, .maxfile = 3 },
+					.out = { .sfile = false,
+						 .ftree = true },
 					.revision = NULL };
 
 static int
 logserver_log_msg_data_null(const struct logserver_msg_data *msg_data)
 {
 	return 0;
+}
+
+static int logserver_openlog(const char *path)
+{
+	int fd = open(path, O_CREAT | O_SYNC | O_RDWR | O_APPEND, 0644);
+
+	struct stat st;
+	if (fstat(fd, &st) != 0)
+		return fd;
+
+	if (st.st_size < logserver_g.log.maxsize)
+		return fd;
+
+	char pattern[PATH_MAX] = { 0 };
+	snprintf(pattern, PATH_MAX, "%s.*.gz", path);
+
+	glob_t glist;
+	int r = glob(pattern, 0, NULL, &glist);
+	if (r != 0 && r != GLOB_NOMATCH)
+		return fd;
+
+	// looking for the newest file
+	size_t max = 0;
+	for (size_t i = 0; i < glist.gl_pathc; ++i) {
+		char str[PATH_MAX] = { 0 };
+		strcpy(str, glist.gl_pathv[i]);
+		str[strlen(str) - 3] = '\0';
+		size_t n = strtoumax(strrchr(str, '.') + 1, NULL, 10);
+		if (n > max)
+			max = n;
+	}
+	// next file
+	++max;
+
+	// only keep maxfile files
+	if ((int)glist.gl_pathc >= logserver_g.log.maxfile) {
+		char delete_path[PATH_MAX] = { 0 };
+		snprintf(delete_path, PATH_MAX, "%s.%zd.gz", path,
+			 max - logserver_g.log.maxfile);
+		pv_fs_path_remove(delete_path, false);
+	}
+
+	globfree(&glist);
+
+	char path_gz[PATH_MAX] = { 0 };
+	snprintf(path_gz, PATH_MAX, "%s.%zd", path, max);
+	pv_fs_file_gzip(path, path_gz);
+
+	ftruncate(fd, 0);
+	lseek(fd, 0, SEEK_SET);
+
+	return fd;
 }
 
 static int
@@ -126,8 +212,6 @@ logserver_log_msg_data_file_tree(const struct logserver_msg_data *msg_data)
 	int ret = -1;
 	char *dup_pathname = NULL;
 	char *fname = NULL;
-	struct stat st;
-	const int MAX_SIZE = 2 * 1024 * 1024;
 	bool source_is_pv = !strncmp(msg_data->platform, PV_PLATFORM_STR,
 				     strlen(PV_PLATFORM_STR));
 
@@ -142,13 +226,9 @@ logserver_log_msg_data_file_tree(const struct logserver_msg_data *msg_data)
 	 */
 	if (pv_fs_mkdir_p(fname, 0755))
 		goto error;
-	log_fd = open(pathname, O_CREAT | O_SYNC | O_RDWR | O_APPEND, 0644);
+
+	log_fd = logserver_openlog(pathname);
 	if (log_fd >= 0) {
-		if (!fstat(log_fd, &st)) {
-			/* Do we need to make a zip out of it?*/
-			if (st.st_size >= MAX_SIZE)
-				ftruncate(log_fd, 0);
-		}
 		if (source_is_pv) {
 			dprintf(log_fd,
 				"[pantavisor] %" PRIu64 " %s\t -- [%s]: %.*s\n",
@@ -180,8 +260,6 @@ logserver_log_msg_data_single_file(const struct logserver_msg_data *msg_data)
 	size_t json_len;
 	int log_fd;
 	char *json = NULL;
-	struct stat st;
-	const int MAX_SIZE = 2 * 1024 * 1024;
 	struct logserver_msg_data msg_data_json_escaped = {
 		.version = msg_data->version,
 		.level = msg_data->level,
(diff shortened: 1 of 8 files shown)


logserver.c
```diff
@@ -529,42 +777,53 @@ void pv_logserver_toggle(struct pantavisor *pv, const char *rev)
...
+	logserver_g.logsock = logserver_open_socket(LOGCTRL_FNAME, true);
+
+	if (logserver_g.out.sfile)
+		logserver_g.fdsock = logserver_open_socket(LOGFD_FNAME, true);
+
+	logserver_g.epfd = epoll_create1(0);
+
+	if (logserver_g.epfd < 0 || logserver_g.logsock < 0 ||
+	    (logserver_g.fdsock < 0 && logserver_g.out.sfile)) {
+		pv_log(DEBUG, "logserver_g epoll_fd = %d", logserver_g.epfd);
```
anibalportero, line 804: should these logs be changed to WARN or ERROR?

logserver.c
```diff
@@ -529,42 +777,53 @@ void pv_logserver_toggle(struct pantavisor *pv, const char *rev)
...
+	}
+
+	logserver_start_service(pv_bootloader_get_rev());
+	pv_log(DEBUG, "started log service with pid %d", (int)logserver_g.pid);
 
 	return 0;
 out:
-	close(logserver_g.sock_fd);
-	close(logserver_g.epoll_fd);
+	close(logserver_g.logsock);
```
anibalportero, line 824: maybe check here is these are > -1 to be closed

logserver.c
```diff
@@ -347,127 +372,350 @@ static int logserver_handle_msg(struct logserver *logserver,
...
-	if (bind(fd, (const struct sockaddr *)&addr, sizeof(addr.sun_path)) ==
-	    -1) {
-		pv_log(ERROR, "unable to bind control socket: %s",
-		       strerror(errno));
-		close(fd);
-		fd = -1;
-		goto out;
-	}
+	// size of sockaddr_un sun_path
+	char path[107];
```
anibalportero, line 685: better use PATH_MAX here if you are going to use it in pv_paths_pv_file

logserver.c
```diff
@@ -347,127 +372,350 @@ static int logserver_handle_msg(struct logserver *logserver,
...
+	int logsock = logserver_g.logsock;
+	int fdsock = logserver_g.fdsock;
+	struct dl_list *tmplst = &logserver_g.tmplst;
+	struct dl_list *fdlst = &logserver_g.fdlst;
+
+	int curfd = -1;
+	for (int i = 0; i < n_events; ++i) {
+		curfd = ev[i].data.fd;
+		if (curfd == logsock || curfd == fdsock) {
+			int fd = logserver_accept_connection(curfd);
```
anibalportero, line 600: maybe do this in different functions (the content of this if, then the content of the else if, then the else)? this kind of code creates a lot of indentation and very big functions like this one. Not necessarily bad but hard to read

logserver.c
```diff
@@ -347,127 +372,350 @@ static int logserver_handle_msg(struct logserver *logserver,
...
-		fd = -1;
-		goto out;
-	}
+	// size of sockaddr_un sun_path
+	char path[107];
+	pv_paths_pv_file(path, PATH_MAX, fname);
+	strncpy(addr.sun_path, path, sizeof(addr.sun_path) - 1);
+
+	if (server) {
+		// sometimes, the socket file still exists after reboot
```
anibalportero, line 690: same here. Not a fan of bool server either for a function that could be two functions as shares little code. Have a look at this thread https://softwareengineering.stackexchange.com/questions/147977/is-it-wrong-to-use-a-boolean-parameter-to-determine-behavior

logserver.c
```diff
@@ -347,127 +372,350 @@ static int logserver_handle_msg(struct logserver *logserver,
...
+			pv_log(ERROR, "error calling epoll_wait: %s",
+			       strerror(errno));
+			return 0;
+		}
+	} while (ready < 0);
+
+	return ready;
+}
+
+static void logserver_consume_data(int fd, bool sub)
```
anibalportero, line 549: Same here

logserver.c
```diff
@@ -678,15 +936,69 @@ void pv_logserver_stop(void)
...
+				 .iov_len = LOGSERVER_MAX_HEADER_LEN };
+	iov[1] = (struct iovec){ .iov_base = src_buf,
+				 .iov_len = LOGSERVER_MAX_HEADER_LEN };
+
+	union {
+		char buf[CMSG_SPACE(sizeof(int))];
+		struct cmsghdr align;
+	} ctrl;
+
+	struct msghdr msg = { .msg_name = NULL,
```
anibalportero, line 981: Not very strong on this, but could we get a send message function for sending messages in general? do you think it would be useful some day?

logserver.c
```diff
@@ -347,127 +372,350 @@ static int logserver_handle_msg(struct logserver *logserver,
...
+	char platform[LOGSERVER_MAX_HEADER_LEN] = { 0 };
+	char src[LOGSERVER_MAX_HEADER_LEN] = { 0 };
+
+	struct iovec iov[2];
+	iov[0] = (struct iovec){ .iov_base = platform,
+				 .iov_len = LOGSERVER_MAX_HEADER_LEN };
+	iov[1] = (struct iovec){ .iov_base = src,
+				 .iov_len = LOGSERVER_MAX_HEADER_LEN };
+
+	struct msghdr msg = { .msg_name = NULL,
```
anibalportero, line 507: Same here, would it be useful to get a receive message function for the future?

logserver.c
```diff
@@ -347,127 +372,350 @@ static int logserver_handle_msg(struct logserver *logserver,
...
-static int logserver_open_socket(const char *path)
+static int logserver_list_add(struct dl_list *tmplst, struct logserver_fd *lfd)
+{
+	if (!lfd)
+		return -1;
+	dl_list_add(tmplst, &lfd->list);
+	return 0;
+}
+
+static struct logserver_fd *logserver_fd_from_list(struct dl_list *l, int fd)
```
anibalportero, line 479: change name to logserver_fetch_fd_from_list or something similar

anibalportero (commented):
Good work! IMO this PR could have been at least two PRs: first a refactor to allow epoll for logserver log sock, then the new fd sock with the lxc changes. Have it in mind for future PRs. I'm sure if we start making smaller PRs it will be easier to develop, test, review and land.

platforms.h
```diff
@@ -74,6 +74,7 @@ struct pv_platform {
 	char *type;
 	char **configs;
 	char *exec;
+	int console_log_fd;
```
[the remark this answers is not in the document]
fortizc: fixed b268875

platforms.c
```diff
@@ -776,13 +791,17 @@ static int pv_platform_stop_loggers(struct pv_platform *p)
 		}
 	}
 
+	close(p->console_log_fd);
```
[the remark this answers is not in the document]
fortizc: Ok, it makes sense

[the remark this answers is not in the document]
fortizc: fixed 0876260

platforms.c
```diff
@@ -724,6 +731,14 @@ int pv_platform_start(struct pv_platform *p)
 			       "Could not start pv_logger for platform %s",
 			       p->name);
 
+	int fd = ctrl->get_console_fd(p, p->data);
+
+	if (fd < 0) {
+		pv_log(ERROR, "Could not get console log fd for %s", p->name);
```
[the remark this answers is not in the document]
fortizc: I agree, fix 790b1b0

logserver.c
fortizc: I will rearrange this function because some fd are more critical than others, for example, the epoll fd. So I will split that `if` and change the log level as appropriate

fortizc: Check 93c1fa8 to see the new implementation

logserver.c
fortizc: Solved in 93c1fa8

logserver.c
fortizc: Yeah, that is on my backlog ;)

logserver.c
fortizc: Mmm.. we have a problem here because the `sun_path` member has only 108 bytes, so if I use `PATH_MAX` I could write out of the bounds of the `sun_path` member.

logserver.c
fortizc: I see... I will refactor this and `logserver_consume_data` thanks for this kind of feedback!

fortizc: fixed d393770

logserver.c
fortizc: fixed a53ce76

logserver.c
anibalportero: Maybe you can apply here the same solution I used here https://github.com/pantavisor/pantavisor/pull/139/files

logserver.c
fortizc: It could be! but I think we can do something like LXC af_unix lib who includes functions to connect, listen, send/receive fd, etc.
https://gitlab.com/pantacor/lxc/-/blob/stable-3.0/src/lxc/af_unix.h

anibalportero:
Looks good to me now :)

platforms.c
```diff
@@ -82,8 +86,9 @@ static struct pv_logger_config plat_logger_config_messages = {
 struct pv_cont_ctrl {
 	char *type;
 	void *(*start)(struct pv_platform *p, const char *rev, char *conf_file,
-		       void *data);
+		       int pipe, void *data);
```
asac, line 89: i dont like the var name "pipe" ... what is this? Also: what is actually in data? Is that something that could hold such logfd?

plugins/pv_lxc.c
```diff
@@ -468,99 +442,14 @@ static void pv_setup_default_log(struct pv_platform *p, struct lxc_container *c,
...
-			pv_log_i = NULL;
-		}
-	}
-out:
-	return pv_log_i;
-}
-
 void *pv_start_container(struct pv_platform *p, const char *rev,
-			 char *conf_file, void *data)
+			 char *conf_file, int logpipe, void *data)
```
asac, line 446: isn't this logfd? if i read pipe i would think of tuple.

plugins/pv_lxc.c
```diff
@@ -468,99 +442,14 @@ static void pv_setup_default_log(struct pv_platform *p, struct lxc_container *c,
...
-/*
- * Use only after loading config.
- * Use threshold as 0 to always truncate.
- */
-static void pv_truncate_lxc_log(struct lxc_container *c,
-				const char *platform_name, off_t threshold,
-				const char *key)
-{
-	char logfile_name[PATH_MAX];
-
```
asac, line 480: this reminds me: do we have a truncation story in the new singlefile and filetree formats?

asac: we have some truncation configs in pantavisor that we would prefer to keep try honoring...

plugins/pv_lxc.c
```diff
@@ -618,14 +506,7 @@ void *pv_start_container(struct pv_platform *p, const char *rev,
...
 		if (pv_lxc_capture_logs_activated()) {
-			__pv_paths_pv_log_plat(log_dir, PATH_MAX,
-					       __pv_get_instance()->state->rev,
-					       p->name);
-			pv_lxc_log.name = LXC_LOG_SUBDIR;
-			pv_lxc_log.lxcpath = strdup(log_dir);
-			if (!pv_lxc_log.lxcpath)
-				goto out_container_init;
-
+			pv_lxc_log.pipe = logpipe;
```
asac, line 509: again logfd both on left and right imo.

plugins/pv_lxc.c
```diff
@@ -681,57 +562,6 @@ void *pv_start_container(struct pv_platform *p, const char *rev,
...
-		{
-			/*
-			 * Change pv_log_i->truncate_size here
-			 * if required.
-			 */
-			pv_log_i = pv_create_lxc_log(p, c, item_config);
-			if (pv_log_i) {
-				const char *truncate_item =
-					pv_log_i->pv_log_get_config_item(
-						item_config, "maxsize");
```
asac, line 702: again, lets check what it takes to keep truncation feature alive...

asac: ... not sure if we had rotation, but if so also worth checking...

plugins/pv_lxc.c
```diff
@@ -344,28 +340,6 @@ static void pv_setup_lxc_container(struct lxc_container *c,
...
-			__pv_paths_pv_log_file(path, PATH_MAX,
-					       __pv_get_instance()->state->rev,
-					       c->name, LXC_CONSOLE_LOG_FNAME);
-			c->set_config_item(c, "lxc.console.logfile", path);
-		}
-	}
-	/*
-	 * Put a hard limit of 2MiB on console file size if one is not defined.
-	 */
-	if (c->get_config_item(c, "lxc.console.size", NULL, 0)) {
```
asac, line 364: if we stop supporting these we need to change config.h and config.c too.... not sure if there is a smart way to honour these in the new world.... if so doing that of course would be amazing if not too much work...

platforms.h
```diff
@@ -77,6 +83,7 @@ struct pv_platform {
 	unsigned long ns_share;
 	void *data;
 	pid_t init_pid;
+	struct pv_platform_log logfd;
```
asac, line 86: isnt this just "log"?

platforms.c
fortizc: Ok, got it, we need a better name `logfd` could be better. 
The `data` parameter is an "out" parameter, returns with the container PID

plugins/pv_lxc.c
fortizc: Ok, I think I should unify the name, logfd will be the name

plugins/pv_lxc.c
fortizc: Right now, logserver uses a constant called MAX_SIZE (lines 155 and 210) to truncate the logs (all generated logs) I know that is not the ideal situation but at least there is no risk of using all the memory. And you are right, we need to support that configuration and I think the better way is inside the logserver.

asac: ok, lets chat about what the current configuration options are on the log front... my memory needs to be reminded i guess as its been a while ...

PR #135, opened 2022-09-01
allow to disable dm_verity through /proc/cmdline pv_verityoff

asac:


state.c
```diff
@@ -1023,7 +1023,12 @@ static char *_pv_state_get_novalidate_list(char *rev)
 	int res;
 	char *cmd = NULL;
 	struct dirent *dp;
-	DIR *volmountdir = opendir("/lib/pv/volmount/verity");
+	DIR *volmountdir;
+
+	if (getenv("pv_verityoff"))
+		return NULL;
+
+	volmountdir = opendir("/lib/pv/volmount/verity");
```
anibalportero, line 1031: IMO this should use pv_paths_lib_volmount to get the prefix if configured

asac: ok, but this is not a regression, rather an oversight in the previous appengine refactor... will land as is and we can change that later....

asac: i can change it if you can post a working line ... but i never used that and dont have time for another retest :)

PR #136, opened 2022-09-06
fix: get upstream json-build resize mem leak fix

anibalportero:
This is not our fix for this problem, which was already merged to master. It is the upstream fix.


asac:
ack

PR #137, opened 2022-09-09
cgroup version detection

anibalportero:
cgroup version is now detected when initializing pantavisor. Based on the /sys/fs/cgroup/ file type and /sys/fs/cgroup/unified/, we determine if the system has the v1, the unified or the v2.

This information is then used from ctrl so sender platform name can be get from v2. It is also used in lxc plugin to substitute lxc.cgroup.devices.allow = a with lxc.cgroup2.devices.allow = a

### DIFF
--- a/ctrl.c
+++ b/ctrl.c
@@ -646,12 +646,50 @@ error:
 	return -1;
 }
 
-static char *pv_ctrl_get_sender_pname(int req_fd)
+static char *pv_ctrl_get_sender_pname_cgroup(FILE *fd)
 {
 	char *pvcg, *pname = NULL;
+	char buf[128];
+
+	while (fgets(buf, 128, fd)) {
+		pvcg = strstr(buf, ":name=pantavisor:/lxc/");
+		if (pvcg) {
+			pvcg += strlen(":name=pantavisor:/lxc/");
+			pvcg[strlen(pvcg) - 1] = '\0';
+			pname = strdup(pvcg);
+			break;
+		}
+	}
+
+	return pname;
+}
+
+static char *pv_ctrl_get_sender_pname_cgroup2(FILE *fd)
+{
+	char *pvcg, *pname = NULL;
+	char buf[128];
+
+	// get container name
+	while (fgets(buf, 128, fd)) {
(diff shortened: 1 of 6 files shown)


ctrl.c
```diff
@@ -1318,8 +1362,23 @@ void pv_ctrl_free_cmd(struct pv_cmd *cmd)
...
 
+	if (statfs("/sys/fs/cgroup/", &fs) < 0) {
+		pv_log(ERROR, "statfs /sys/fs/cgroup/ failed: %s",
+		       strerror(errno));
+		return -1;
+	}
+
+	if (fs.f_type == CGROUP2_SUPER_MAGIC) {
+		pv_log(INFO, "system is using cgroup v2");
+		pv->cgroupv = 2;
```
asac, line 1376: we should here really also determine mixed mode.... mixed is if cgroup2 is mounted on /sys/fs/cgroup//unified.... mixed will allow us later to better filter/generate the cgroup2 / cgroup1 rules i guess

init.c
```diff
@@ -113,6 +113,12 @@ static int mount_cgroups()
 {
 	int ret;
 
+	mkdir("/sys/fs/cgroup/pantavisor", 0555);
```
asac, line 116: we should only do this in v1 mode imo....

anibalportero: This results in /proc/PID/cgroup that look like this:

```
1:name=pantavisor:/
0::/lxc/pvr-sdk
```

With this, we can check in ctrl if the PID has "name=pantavisor". This adds an extra layer of security to differentiate between pantavisor and non-pantavisor lxc containers.

Also, it prevents from executing more than one instances of Pantavisor from an early stage.

ctrl.c
anibalportero: Added this. Now we detect if in v2 with as with the previous version. If not in v2, we check if the unified file exists or not to determine whether we are in unified or v1.

init.c
anibalportero: done

ctrl.c
anibalportero: done

PR #138, opened 2022-09-12
make dropbear use the dynamic paths; enable ssh for appengine

anibalportero:


asac:
what happened to this one?

anibalportero:
> what happened to this one?

Nothing in particular. It was just temporarily abandoned to focus on other tasks, but I still think we need something like this to get ssh on appengine.

PR #139, opened 2022-09-13
fix: bind not being used as in the man example

anibalportero:
There were a number of misuses of the bind function in both logserver and ctrl:
* the third argument of bind should be the size of the addr struct, not the addr.sun_path size ([source man page](https://www.man7.org/linux/man-pages/man2/bind.2.html))
* addr.sun_path not properly initialized to 0 to avoid buffer overflows  ([source man page](https://www.man7.org/linux/man-pages/man2/bind.2.html))

### DIFF
--- a/ctrl.c
+++ b/ctrl.c
@@ -111,7 +111,7 @@ static const char *pv_ctrl_string_http_status_code(pv_http_status_code_t code)
 	return strings[code];
 }
 
-static int pv_ctrl_socket_open(char *path)
+static int pv_ctrl_socket_open()
 {
 	int fd;
 	struct sockaddr_un addr;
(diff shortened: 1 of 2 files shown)


ctrl.c
```diff
@@ -1318,9 +1319,9 @@ void pv_ctrl_free_cmd(struct pv_cmd *cmd)
 static int pv_ctrl_init(struct pv_init *this)
 {
 	struct pantavisor *pv = pv_get_instance();
-	char path[PATH_MAX];
+	char path[107];
```
asac, line 1322: why exactly 107?

asac:
@anibalportero please add the manpage example to the commit msg (and say which manpage that is from)

ctrl.c
anibalportero: is the sun_path size. Added that as a comment

PR #140, opened 2022-09-13
Remove call to pv_logserver_stop when is a non-reboot update

fortizc:


anibalportero:
Does this work? Is logserver changing to the proper /pv/logs/revision path when doing a non-reboot transition?

asac:
@anibalportero no it does not. needs urgent fix.

PR #143, opened 2022-09-14
fix: groups not cleaning old platform references during non-reboot transitions

anibalportero:
Currently, in non-reboot transitions, Pantavisor keeps the current state running for the new revision, but transfers the new/updated platforms from the pending state to the running one.This does not affect groups, as they are not changed whatsoever during non-reboot updates. We still have to re-link the platform reference list from the pending state to the current one.

Before this fix, Pantavisor was just re-linking the platforms belonging to the new revision, but the old ones belonging to platforms that were deleted or platforms that were modified were still there. This was causing Pantavisor to keep groups with lists of platform references that pointed to deleted structs etc.

With this fix, the list of references for each group are deleted during transitions, so the platforms belonging to the new revision can be cleanly re-linked to the groups.

### DIFF
--- a/group.c
+++ b/group.c
@@ -44,7 +44,7 @@ struct pv_group *pv_group_new(char *name)
 	return g;
 }
 
-static void pv_group_empty_platform_refs(struct pv_group *g)
+void pv_group_empty_platform_refs(struct pv_group *g)
 {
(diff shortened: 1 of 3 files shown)


asac:
+1

PR #145, opened 2022-09-21
fix: runlevel not having the same status as group in run.json

anibalportero:
With this PR, we enable any string in run.json "runlevel", making it equivalent to "group". In case the two of them are configured and their values differ, the revision will be set as WONTGO.

### DIFF
--- a/parser/parser_system1.c
+++ b/parser/parser_system1.c
@@ -824,34 +824,6 @@ static int do_action_for_type(struct json_key_action *jka, char *value)
 	return 0;
 }
 
-static int do_action_for_runlevel(struct json_key_action *jka, char *value)
-{
-	struct pv_group *g;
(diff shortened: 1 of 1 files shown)


asac:
i would also be fine to just klll group adn wait for state v3 before doing such vocabulary change

anibalportero:
As runlevel only supports data root, platform or app values, open that to other possible values would make the new containers incompatible with old devices. I think that's a good enough reason to keep runlevel the way it is and start using group, which has more flexibility as it allows any value

anibalportero:
As runlevel only supports data root, platform or app values, open that to other possible values would make the new containers incompatible with old devices. I think that's a good enough reason to keep runlevel the way it is and start using group, which has more flexibility as it allows any value

PR #153, opened 2022-11-04
close logserver fds before killing its pid

anibalportero:
A bunch of small changes I did while testing appengine shuts down before WAIT state is reached. List of changes:
* ctrl: unlink before binding to remove existing pv-ctrl sockets
* logserver: merge close and stop functions
* logserver: add fd closing before stopping the PID, which was causing a freeze
* pantavisor: stop logserver after we stop ph_logger, platforms and pvctrl, but before unmount and free up memory
* ph_logger: improve logs when stopping services


asac:
does this have implications on how system mode is doing reboots and revision transition?

anibalportero:
This changes things in the part of shutdown that we only execute in appengine

PR #155, opened 2022-11-10
improve scripts/volmount/crypt/crypt to be more robust against corruption

asac:
1. we will move to explicit .init_done marker file that we only place if whole do_crypt_init function has succeeded with changes synched to disk
2. we only do losetup logic for dm_type caam
3. we create a .goodbak copy of the image file every time we successfully finish with do_mount_disk; we are extra careful and create a .prebak copy of the image file before the mounting happens.
4. we make do_mount_disk try the following before giving up: - mount crypt disk as usual - if fails run an fsk.ext4 -y on that /dev/mapper/ device - if that fails we put back the .goodbak file (if such exist) at code we do this through recursive invocation of the do_mount_disk function with a 4th parameter 'specialrun' mode

### DIFF
--- a/scripts/volmount/crypt/crypt
+++ b/scripts/volmount/crypt/crypt
@@ -62,16 +62,13 @@ do_crypt_init()
 		cat /data/caam/$keyname | /bin/keyctl padd logon logkey: @u
 		[ $? -ne 0 ] && echo "ERROR: CAAM: Unable to load key" && exit 1
 	elif [ $dm_type == "dcp" ]; then
-		key_value=`head /dev/urandom | tr -dc A-Za-z0-9 | head -c 32`
-		keyfile=`/bin/mktemp`
-		echo $key_value > $keyfile
-		/lib/pv/volmount/crypt/dcp-tool enc $keyfile $dir/$keyname.bb 64
-		[ $? -ne 0 ] && echo "ERROR: DCP: Unable to create black blob" && exit 1
-		rm -f $keyfile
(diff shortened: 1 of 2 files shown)


asac:
@parthitce please help review and test corner cases on this one.

scripts/volmount/crypt/crypt
```diff
@@ -169,18 +175,52 @@ do_mount_disk()
...
+		sync
+		return 0
+	fi
+
+	echo "ERROR: could not setup dm-crypt volume"
+	if [ "$specialrun" == "fsck" ] && [ -f $dir/$img_file.goodbak ]; then
+		echo "Specialrun $specialrun did not succeed ... trying from goodbackup"
+		cp -f $dir/$img_file.goodbak $dir/$img_file
+		/usr/sbin/cryptsetup close $device_name || true
+		if ! do_mount_disk "$dm_type" "$path" "$mntpath" bak; then
```
parthitce, line 201: Will it be endless recursion here?

asac: it passes the "bak" argument at last position which should be the end run ... maybe check. i triggered that case manually as  i couldnt get fsck to not produce something mountable (but there was data loss which is not taht great).

asac:
we dropped the dcp-tool and replaced it with a plugin that vendors can put in place ... called x--pv-dcp-tool.

Also changed that only first successful boot does a goodbak so we dont need to inspect post mount result if there was important data loss.

PR #157, opened 2022-11-15
Add new tsh library

fortizc:
Improve TSH library
* Reuse code for all functions
* Replace `select` and friends by `epoll`
* Keep the API to avoid compatibility issues

### DIFF
--- a/atom.mk
+++ b/atom.mk
@@ -111,7 +111,7 @@ LOCAL_C_INCLUDES := $(LOCAL_PATH)
 
 LOCAL_SRC_FILES := utils/tsh.test.c \
 			utils/tsh.c \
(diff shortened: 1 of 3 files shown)


asac:
I would like to have the out buffers rather grow to  a MAXSIZE then having to prealloc always a big chunk on stack.... what do you think?

fortizc:
> I would like to have the out buffers rather grow to a MAXSIZE then having to prealloc always a big chunk on stack.... what do you think?

Ok, but that means an API change, right? because if the out buffers grow, then the size parameter is an output parameter, although we always return a null terminated string and forget about the size. Any case, we would be changing the API

PR #159, opened 2022-11-15
Implement loglevel on subscribed fds

fortizc:
* Add loglevel to `logserver_subscribe_fd`
* Set INFO as default loglovel for containers fds


asac:
lgtm... if tested land it....

PR #163, opened 2022-11-22
Add initial qos story for system and containers

asac:
Add initial oom score adjustments to ensure that on pantavisor system pantavisor itself and the log facilities get killed last by oom handler

### DIFF
--- a/config.c
+++ b/config.c
@@ -445,6 +445,11 @@ static int pv_config_load_file(char *path, struct pantavisor_config *config)
 	config_iterate_items_prefix(&config_list, config_sysctl_apply,
 				    "sysctl.", NULL);
 
+	config->vm.oom_qos_besteffort = config_get_value_int(
+		&config_list, "vm.oom_qos_besteffort", 1000);
+	config->vm.oom_qos_guaranteed = config_get_value_int(
+		&config_list, "vm.oom_qos_guaranteed", -997);
+
 	config_clear_items(&config_list);
 
 	return 0;
@@ -724,6 +729,10 @@ void pv_config_override_value(const char *key, const char *value)
 		pv->config.metadata.devmeta_interval = atoi(value);
 	else if (!strcmp(key, "metadata.usrmeta.interval"))
 		pv->config.metadata.usrmeta_interval = atoi(value);
+	else if (!strcmp(key, "vm.oom_qos_besteffort"))
+		pv->config.vm.oom_qos_besteffort = atoi(value);
+	else if (!strcmp(key, "vm.oom_qos_guaranteed"))
+		pv->config.vm.oom_qos_guaranteed = atoi(value);
 	else if (!strcmp(key, "debug.ssh"))
 		pv->config.debug.ssh = atoi(value);
 }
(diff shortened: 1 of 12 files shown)


asac:
pushed new revision that supports qos_policy: "guaranteed, besteffort and burstable"; also sets system pids to guaranateed.

asac:
TODO:
 * make actual values for each level configurable
 * make setting for system pids individually tunable
 * make dropbear get prio
 * take a stab at making burstable more like k8s by looking at mem settings.

asac:
i fixed all TODOS except the burstable impl is still dumb... would like to land as is...

asac:
think memleak still with ```p->qos_policy_oom_score_adj = strdup(score);```

config.h
```diff
@@ -186,6 +186,11 @@ struct pantavisor_metadata {
 	int usrmeta_interval;
 };
 
+struct pantavisor_vm {
+	int oom_qos_besteffort;
```
anibalportero, line 190: whould be nice to have these in pv_config_get_json and pv_config_print too

config.c
```diff
@@ -445,6 +445,11 @@ static int pv_config_load_file(char *path, struct pantavisor_config *config)
 	config_iterate_items_prefix(&config_list, config_sysctl_apply,
 				    "sysctl.", NULL);
 
+	config->vm.oom_qos_besteffort = config_get_value_int(
+		&config_list, "vm.oom_qos_besteffort", 1000);
```
anibalportero, line 449: I don't like the names very much, but I cannot think of something better either :)

init.c
```diff
@@ -466,6 +466,21 @@ int main(int argc, char *argv[])
 
 	// create pv thread
 	pv_pid = fork();
+	if (pv_pid > 0 && getuid() == 0) {
```
anibalportero, line 469: aren't we doing the same in this if and else bodies?

parser/parser_system1.c
```diff
@@ -1118,6 +1118,36 @@ static int do_action_for_drivers(struct json_key_action *jka, char *value)
...
+{
+	if (pv_str_matches(value, len, "guaranteed", strlen("guaranteed")))
+		return QOS_GUARANTEED;
+	else if (pv_str_matches(value, len, "besteffort", strlen("besteffort")))
+		return QOS_BESTEFFORT;
+	else if (pv_str_matches(value, len, "burstable", strlen("burstable")))
+		return QOS_BURSTABLE;
+
+	pv_log(ERROR, "invalid qos policy '%s'", value);
+	return QOS_NONE;
```
anibalportero, line 1131: is QOS_NONE the same as QOS_UNKNOWN?

anibalportero: also, bear in mind that QOS_NONE won't fail in this case

platforms.c
```diff
@@ -474,6 +474,16 @@ static int __start_pvlogger_for_platform(struct pv_platform *platform,
 	if (!pid) {
 		char namespace[64];
 		int ns_fd = -1;
+
+		if (getuid() == 0) {
```
anibalportero, line 479: trying to undertand this. What are the getuid around doing?

asac: @anibalportero only for containers that get started as root we do this... if we are not root we cannot adjust oom score and it woudl fail.

parser/parser_system1.c
asac: QOS_NONE is unset, while UNKNOWN is not parsed (yet).

config.c
asac: we can add a . in the middle, but then i dont know the benefit :)

config.h
asac: sure lets add there....

init.c
asac: no the other else is for the child process (e.g. pv_pid == 0) not > 0

anibalportero: well, I mean what's inside of the body of the if and the else is the same
```
	if (pv_system_oom_adjust(
			    pv_config_get_vm_oom_qos_guaranteed())) {
			printf("ERROR: oom adjust failed: '%s'\n",
			       strerror(errno));
			return -1;
```

asac:
needs docs/ update for new config keys.

PR #165, opened 2022-11-22
dcp: use decrypted plain text key for dm-crypt

parthitce:
decrypt the with dcp-tool decrypt and consume as plain text with cryptsetup plain.


asac:
this is outdated.

PR #175, opened 2022-11-30
Add dmesg on pantavisor log

fortizc:
* Add `log.capture.dmesg` to activate or deactivate (1 or 0) this feature
* Add a fd from `/dev/kmsg` to logserver subscription

### DIFF
--- a/logserver.c
+++ b/logserver.c
@@ -929,6 +929,16 @@ static int logserver_open_server_socket(const char *fname)
 	return fd;
 }
 
+static void logserver_drop_fds(struct dl_list *lst)
+{
+	struct logserver_fd *it, *tmp;
+	dl_list_for_each_safe(it, tmp, lst, struct logserver_fd, list)
+	{
+		logserver_remove_fd(it->fd);
+		logserver_fd_free(it);
+	}
+}
+
 static pid_t logserver_start_service(const char *revision)
 {
 	logserver_g.pid = fork();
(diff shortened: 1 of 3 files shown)


logserver.c
```diff
@@ -983,6 +983,20 @@ void pv_logserver_toggle(struct pantavisor *pv, const char *rev)
 	}
 }
 
+static void logserver_capture_dmesg()
+{
+	errno = 0;
+	int fd = open("/dev/kmsg", O_RDONLY | O_NONBLOCK);
```
anibalportero, line 989: this is a good example for me to learn how this works as it is pretty simple :) Where are we closing this?

logserver.c
```diff
@@ -983,6 +983,20 @@ void pv_logserver_toggle(struct pantavisor *pv, const char *rev)
...
+	if (fd < 0) {
+		pv_log(DEBUG, "could not susbcribe dmesg to logserver %s",
```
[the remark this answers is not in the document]
asac: yeah this is WARN if its something a setting has explicitely requested.

logserver.c
asac: i am not sure if we have to close ita fter sending it to logserver ... logserver definitely closes the fd by itself so .... lets try really and be sure we use this correctly indeed.

logserver.c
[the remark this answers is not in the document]
fortizc: Done

logserver.c
fortizc: Currently, the system relies on two ways for closing the fds, the first is the unsubscribe function that function and the second is the detection of fds in bad states. So @anibalportero is right this fd in particular never be close, maybe we will need a function to close and unsubscribe all fds when we want to shut down the system (maybe only in the pv_logserver_stop() function?)

PR #176, opened 2022-12-01
make clean shutdown

asac:
* unmount disks
* degrade logs instead of killing; log to stdout after umount of /storage
* reboot at last possible moment with just logserver running
* improve error return code for pv_state_stop_force
* manually kick wdt after setting timeout

### DIFF
--- a/pantavisor.c
+++ b/pantavisor.c
@@ -853,6 +853,8 @@ static pv_state_t pv_shutdown(struct pantavisor *pv, shutdown_type_t t)
 	if (!pv)
 		return PV_STATE_EXIT;
 
+	init_mode_t initmode = pv_config_get_system_init_mode();
+
 	pv_log(INFO, "preparing '%s'...", shutdown_type_string(t));
 	if (pv_config_get_system_init_mode() == IM_APPENGINE)
 		pv_log(INFO,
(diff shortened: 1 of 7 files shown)


pantavisor.c
```diff
@@ -872,25 +874,34 @@ static pv_state_t pv_shutdown(struct pantavisor *pv, shutdown_type_t t)
...
 	pv_init_umount();
 
 	// free up memory
 	pv_bootloader_remove();
-	pv_remove(pv);
+
+	// at this point, we can shutdown if not in appengine
+	if (initmode != IM_APPENGINE) {
+		pv_log(INFO, "shutdown complete, rebooting in 2 second ...");
+		sleep(2);
```
anibalportero, line 897: out of curiosity, why do we need the sleep here? Also, @fortizc is it worth it to call pv_logserver_stop in this case too?

state.c
```diff
@@ -627,16 +627,16 @@ int pv_state_stop_force(struct pv_state *s)
 
 	if (!pv_state_check_all_stopped(s)) {
 		pv_state_force_stop(s);
-		ret = -1;
+		ret |= -2;
```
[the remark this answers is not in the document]
asac: nowhere, but thought would be nice to preserve the error codes to be different depending on what test failed.

pantavisor.c
asac: we can call stop also in system reboot case, but we want to see even the last of messages that we can retrieve like a dmesg output at the end telling us that it will reboot etc.... so i thought keeping a single process that has no resources lockd to disk resources should be fine.

PR #186, opened 2022-12-19
avoid loading pantahub.config in local mode; stop supporting deprecated ph.config keys

anibalportero:
This PR moves the loading of pantahub.config out of the early init so we can decide whether to load it or not based on Pantavisor being in remote or local mode. The idea is to only load the config file after bootup or transition once. Then, it is saved only after registration.

Because of this, devices might not load pantahub.config when in local mode, so we need to move the deprecated keys from pantahub.config to pantavisor.config to make this change complete.

List of changes:
* config: stop supporting deprecated pantahub.config keys
* config: print all config keys after log is initialized
* config: avoid loading pantahub.config on early init if remote is off
* config: improve logs for load/save pantahub.config
* log: move config print to config

### DIFF
--- a/config.c
+++ b/config.c
@@ -484,32 +484,13 @@ static int pv_config_load_creds_from_file(char *path,
 	config->factory.autotok =
 		config_get_value_string(&config_list, "factory.autotok", NULL);
 
-	config_override_value_bool(&config_list, "updater.keep_factory",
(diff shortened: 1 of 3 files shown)


asac:
would like to cut 019 today/tomorrow and then we can land such a change without feeling too bad :)

PR #187, opened 2022-12-19
stop replacing good objects in ctrl PUT /objects

asac:
* we check hashsum of target object
 * if exist we dont replace it with a new upload
 * this reduces unecessary inode shuffling and disk duplication

### DIFF
--- a/ctrl.c
+++ b/ctrl.c
@@ -924,8 +924,16 @@ static struct pv_cmd *pv_ctrl_process_endpoint_and_reply(
 			if (pv_ctrl_process_put_file(req_fd, content_length,
 						     file_path_tmp) < 0)
 				goto out;
-			if (pv_storage_validate_file_checksum(file_path_tmp,
-							      file_name) < 0) {
+			if (pv_fs_path_exist(file_path) &&
+			    !pv_storage_validate_file_checksum(file_path,
+							       file_name)) {
+				pv_log(WARN,
+				       "object %s already exists and is valid; discarding new object upload",
+				       file_path_tmp);
+				pv_fs_path_remove(file_path_tmp, false);
+
+			} else if (pv_storage_validate_file_checksum(
+					   file_path_tmp, file_name) < 0) {
 				pv_log(WARN, "object %s has bad checksum",
 				       file_path_tmp);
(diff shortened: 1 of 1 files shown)


ctrl.c
```diff
@@ -924,25 +924,35 @@ static struct pv_cmd *pv_ctrl_process_endpoint_and_reply(
 			if (pv_ctrl_process_put_file(req_fd, content_length,
 						     file_path_tmp) < 0)
 				goto out;
-			if (pv_storage_validate_file_checksum(file_path_tmp,
-							      file_name) < 0) {
+			if (pv_fs_path_exist(file_path) &&
```
anibalportero, line 927: shouldn't we do this before pv_ctrl_process_put_file and return a 409 in case it exists and checksum is good? if not good we can remove and proceed with pv_ctrl_process_put_file. That way we would avoid unnecessary file transfers

ctrl.c
```diff
@@ -924,25 +924,35 @@ static struct pv_cmd *pv_ctrl_process_endpoint_and_reply(
...
-			pv_log(DEBUG, "renaming %s to %s", file_path_tmp,
-			       file_path);
-			if (pv_fs_path_rename(file_path_tmp, file_path) < 0) {
-				pv_log(ERROR, "could not rename: %s",
-				       strerror(errno));
-				pv_ctrl_write_error_response(
-					req_fd, HTTP_STATUS_ERROR,
-					"Cannot rename object");
-				goto out;
+			} else {
```
anibalportero, line 944: this block could go out of the else as we have a goto in the previous if and so we avoid unnecessary indentation

ctrl.c
asac: sure thought about it, but dont want to change behaviour .... there are clients written against this that then suddenly start erroring etc.... so I chose not to do that.

ctrl.c
asac: the first if has no goto ... so its needed.

asac: first we want to send also OK response hence this else....

asac: maybe worth checking if we delete the _tmp file in that case? do we?

PR #190, opened 2022-12-22
load config policies after pantavisor.config

anibalportero:
Pantavisor can now load different policies at boot time. A policy is a configuration file following the same format as pantavisor.config. The files are located in /etc/pantavisor/policies/, so they can be added to each platform from vendor/stepskel/etc/pantavisor/policies/. The policy is set in the first stages of configuration (pantavisor.config or cmdline) and then loaded before the latest stages (initrdconfig or metadata config).

For example, if we wanted to load this policy from vendor:

```
vendor/stepskel/etc/pantavisor/policies/prod.config
```

We would have to set this key in cmdline:

```
pv_policy=prod
```

Another thing that would work is to set it in pantavisor.config

```
policy=prod
```

List of changes:
* config: add new "policy" key to pantavisor.config, overridable with pv_ suffix in cmdline
* config: new "policy" key is a string without '/' characters
* path: add new dynamic path for policies

### DIFF
--- a/config.c
+++ b/config.c
@@ -90,6 +90,24 @@ static bool config_get_value_bool(struct dl_list *config_list, char *key,
 	return value;
 }
 
+static char *config_get_value_policy(struct dl_list *config_list, char *key,
+				     char *default_value)
+{
+	char *item = config_get_value(config_list, key);
+
+	if (!item)
+		item = default_value;
+
+	if (!item || !strlen(item))
+		return NULL;
+
+	// policy does not allow paths
+	if (strchr(item, '/'))
+		return NULL;
+
+	return strdup(item);
+}
+
 static int config_get_value_init_mode(struct dl_list *config_list, char *key,
 				      init_mode_t default_value)
 {
@@ -250,8 +268,24 @@ static int config_sysctl_apply(char *key, char *value, void *opaque)
 	return 0;
 }
 
-static int pv_config_load_config_from_file(char *path,
-					   struct pantavisor_config *config)
+static int pv_config_load_policy(const char *policy,
+				 struct dl_list *config_list)
+{
+	char path[PATH_MAX];
+
+	if (!policy)
+		return 0;
+
+	pv_paths_etc_policy_file(path, PATH_MAX, policy);
+	if (load_key_value_file(path, config_list) < 0) {
+		printf("FATAL: unable to parse %s\n", path);
+		return -1;
+	}
+
+	return 0;
+}
+
+static int pv_config_load_file(char *path, struct pantavisor_config *config)
 {
 	DEFINE_DL_LIST(config_list);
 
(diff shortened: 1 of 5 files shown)


config.c
```diff
@@ -1316,11 +1345,19 @@ int pv_config_init(char *path)
...
 	printf("DEBUG: loading config from %s\n", path);
-	if (pv_config_load_config_from_file(path, &pv->config) < 0) {
+	if (pv_config_load_file(path, &pv->config) < 0) {
 		printf("FATAL: unable to parse %s\n", path);
 		return -1;
 	}
 
+	const char *policy = pv_config_get_policy();
+
+	printf("DEBUG: loading policy '%s'\n", policy);
```
asac, line 1355: are we sure we cannot use pv_log at this point yet?

paths.c
```diff
@@ -370,13 +370,20 @@ void pv_paths_lib_lxc_lxcpath(char *buf, size_t size)
 }
 
 #define PV_ETC_PATHF "%s/%s"
+#define PV_ETC_POLICY_PATHF "%s/pantavisor/policy/%s"
```
asac, line 373: lets name it policies

paths.c
```diff
@@ -370,13 +370,20 @@ void pv_paths_lib_lxc_lxcpath(char *buf, size_t size)
...
 
 void pv_paths_etc_file(char *buf, size_t size, const char *name)
 {
 	SNPRINTF_WTRUNC(buf, size, PV_ETC_PATHF, pv_config_get_system_etcdir(),
```
asac, line 377: lets normalize the filename so someone passing in ../../some/other/unexpected/path wont be able to trick the parser trying to load files that were not intended to be used as policies.

one idea is to send the name through a mangler that replaces all "/" with "-"

config.c
anibalportero: this is previous to logserver initialization

paths.c
anibalportero: done

paths.c
anibalportero: this has further implications. There are a bunch of paths that can be configured in pantavisor.config and/or any other policy config file already (almost every path in pantavisor is configurable since appengine). Should we shield those too? and how?

config.c
```diff
@@ -250,8 +268,25 @@ static int config_sysctl_apply(char *key, char *value, void *opaque)
...
+static int pv_config_load_policy(const char *policy,
+				 struct dl_list *config_list)
+{
+	char path[PATH_MAX];
+
+	if (!policy)
+		return 0;
+
+	pv_paths_etc_policy_file(path, PATH_MAX, policy);
+	printf("DEBUG: loading policy from %s\n", path);
```
asac, line 280: do we really want to print that always regardless of log level? can we not use pv_log here?

config.c
```diff
@@ -1316,7 +1365,7 @@ int pv_config_init(char *path)
 		path = PV_PANTAVISOR_CONFIG_PATH;
 
 	printf("DEBUG: loading config from %s\n", path);
```
asac, line 1367: same for this printf

config.c
anibalportero: if I change it to pv_log it doesn't work for some reason. It's supposed to print to stdout if logserver is not initialized, but it doesn't

PR #194, opened 2023-01-09
make truststore configurable through securetboot.truststore

asac:
* move trust certs and pub keys into folder /etc/pantavisor/pvs/trust/

### DIFF
--- a/config.c
+++ b/config.c
@@ -420,6 +420,9 @@ static int pv_config_load_file(char *path, struct pantavisor_config *config)
 	config->libthttp.loglevel =
 		config_get_value_int(&config_list, "libthttp.log.level", 3);
 
+	config->libthttp.certdir = config_get_value_string(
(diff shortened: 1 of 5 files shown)


config.c
```diff
@@ -420,6 +420,9 @@ static int pv_config_load_file(char *path, struct pantavisor_config *config)
 	config->libthttp.loglevel =
 		config_get_value_int(&config_list, "libthttp.log.level", 3);
 
+	config->libthttp.certdir = config_get_value_string(
```
anibalportero, line 423: we have to free certdir and trustore in pv_config_free

asac: We have:

if (pv->config.libthttp.certdir)
		free(pv->config.libthttp.certdir);

Is that not enougH?

anibalportero: sorry, didn't see those

PR #195, opened 2023-01-09
set empty uboot var in boot.txt instead of deleting it; fix ssh enabling

anibalportero:
This PR modifies uboot code so we set an empty var when unseting (only used for pv_rev=). It also fixes ssh activation, which was being set to true when the arg "splash" was not present. Now, besides being configurable, one can enable debug ssh and shell with the "pv_debug" arg.


asac:
lgtm. lets land it

PR #199, opened 2023-01-13
Fix timer start

fortizc:
Now group timers start only after each group finish its start process, that's mean that each group has the complete timeout to start.

### DIFF
--- a/group.c
+++ b/group.c
@@ -21,6 +21,7 @@
  */
 #include <string.h>
 #include <stdio.h>
+#include <stdbool.h>
 
 #include "group.h"
 
@@ -39,7 +40,8 @@ struct pv_group *pv_group_new(char *name, int timeout, plat_status_t status,
 	g = calloc(1, sizeof(struct pv_group));
 	if (g) {
 		g->name = strdup(name);
-		g->timeout = timeout;
+		g->timeout.limit = timeout;
+		g->timeout.started = false;
 		g->default_status_goal = status;
 		g->default_restart_policy = restart;
 		dl_list_init(&g->platform_refs);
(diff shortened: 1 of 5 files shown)


group.c
```diff
@@ -39,7 +40,8 @@ struct pv_group *pv_group_new(char *name, int timeout, plat_status_t status,
 	g = calloc(1, sizeof(struct pv_group));
 	if (g) {
 		g->name = strdup(name);
-		g->timeout = timeout;
+		g->timeout.limit = timeout;
+		g->timeout.started = false;
```
anibalportero, line 44: I don't like this very much. why not putting started outside? in the end what was started is the group itself right? Maybe it's not even needed, as you can check if the timeout was initialized already?

state.c
```diff
@@ -484,53 +484,75 @@ static int pv_state_start_platform(struct pv_state *s, struct pv_platform *p)
 int pv_state_run(struct pv_state *s)
 {
 	int ret = 0;
-	struct pv_platform *p, *tmp_p;
-
-	dl_list_for_each_safe(p, tmp_p, &s->platforms, struct pv_platform, list)
+	bool stop = false;
+	struct pv_group *g, *tmp_g;
+	dl_list_for_each_safe(g, tmp_g, &s->groups, struct pv_group, list)
```
anibalportero, line 489: why do we need to iterate over groups here? can't we just move the login to pv_state_start_platform? and just initialized the timer of the group belonging to such platform if it wasn't already initialized when we start a platform? does this work for both update and non update bootup?

group.c
fortizc: > I don't like this very much. why not putting started outside? in the end what was started is the group itself right?
Ok, it could be outside, it's only I think it could be better to group those elements, but that not a problem. 

> Maybe it's not even needed, as you can check if the timeout was initialized already?
I don't see any way to do that, if it's exists, maybe this is the opportunity to add a function or a field in `timer_state`

state.c
fortizc: This is the important change here... maybe it's only a misunderstood but AFAIK the groups has a dependency relationship between them in descendant order, i.e., the groups named first in the device.json must be processed before. Now based on that (and that could be the root of all the problem) you cannot iterate over the platforms, because the system not add them in the right order.

anibalportero: That is not needed here. The group order works without your change, we are not introducing that feature here, as it is already working in master. It is implemented in pv_state_check_goals, where we iterate over groups and break if the group of the platform that we are checking are the same. This means we only check the previous groups, as they are ordered in memory. It's not beautiful, but adding another layer here is not going to help either.

So here, you only need to do what I mention in the comment. Check if the timer has been initialized when we start the platform.

group.c
anibalportero: If it's too much hassle to check if a timer has been initialized, just use the variable, but not inside of the struct. This is very similar to plat_status_t. So maybe do something similar to what we do there.

PR #206, opened 2023-01-20
Modular logserver outputs

fortizc:
Creates a new system to allow modular outputs in logserver. The new system uses a base struct to define an output and internally logserver use a list to store the needed outputs (following the configuration).

The current PR includes:
* define a base structure for new outputs (logserver_out)
* rewrite all old outputs using the new system
* creates a small "utils" lib for common functions

### DIFF
--- a/logserver/logserver_utils.c
+++ b/logserver/logserver_utils.c
@@ -1,0 +1,172 @@
+/*
+ * Copyright (c) 2023 Pantacor Ltd.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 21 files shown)


asac:
remember to add copyright everywhere to good content :)

fortizc:
> remember to add copyright everywhere to good content :)

Right! I forgot

logserver/logserver_utils.c
```diff
@@ -0,0 +1,151 @@
...
+			      js.plat, pv_log_level_name(js.lvl), js.src,
+			      (int)strlen(js.data.buf),
+			      js.data.buf) +
+		     1; // 0 byte
+
+	char *json = calloc(1, len);
+	if (!json)
+		return NULL;
+
+	snprintf(json, len, LOGSERVER_JSON_FORMAT, js.tsec, js.tnano, js.plat,
```
anibalportero, line 126: I think we should use the pv_json_ser functions here. We can leave it for another PR if this is just moving things around

logserver/logserver_utils.h
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+#ifndef LOGSERVER_UTILS_H
+#define LOGSERVER_UTILS_H
+
+#include "logserver_out.h"
+
+#include <stdbool.h>
+
+int logserver_open_logfile(const char *path);
```
anibalportero, line 30: we should stick to the logserver_utils_ prefix as you have done in the rest of functions in .h files

anibalportero (commented):
As this change is old, you should  go through all the changes suffered by logserver.h and logserver.c since you opened it. I think not all of them are here. For example, compare pv_logserver_stop in master with yours and you'll see what I mean

PR #207, opened 2023-01-20
feature: disable/enable ssh server through user-meta

anibalportero:
This PR introduces "debug.ssh" to user meta, so dropbear can be started and stopped on run time. It also refactors all debug code to debug.h and debug.c.

List of changes:
* config: override debug.ssh from step config and user-meta
* debug: module with ssh and shell code
* debug: store pids for dropbear and shell session
* debug: functions to start and wait for shell
* debug: functions to start, stop and toggle ssh
* init: all debug related code moved to debug module
* pantavisor: all debug related code moved to debug module

### DIFF
--- a/volumes.c
+++ b/volumes.c
@@ -313,6 +313,7 @@ int pv_volume_mount(struct pv_volume *v)
 	int loop_fd = -1, file_fd = -1;
 	struct pantavisor *pv = pv_get_instance();
 	struct pv_state *s = pv->state;
+	struct pv_disk *d = v->disk;
 	char path[PATH_MAX], mntpoint[PATH_MAX];
 	char *fstype;
 	char *umount_cmd = NULL;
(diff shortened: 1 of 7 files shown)


asac:
for me this works...

asac:
land?

asac:
problem: with this patch not having debug.ssh set in usrmeta changes default from 1 to 0... seems unwanted side effect...

asac:
> problem: with this patch not having debug.ssh set in usrmeta changes default from 1 to 0... seems unwanted side effect...

sorry false alert... couldnt repro

volumes.c
```diff
@@ -313,6 +313,7 @@ int pv_volume_mount(struct pv_volume *v)
 	int loop_fd = -1, file_fd = -1;
 	struct pantavisor *pv = pv_get_instance();
 	struct pv_state *s = pv->state;
+	struct pv_disk *d = v->disk;
```
anibalportero, line 316: Why these changes in volumes.c? I think this is something we changed after creating this PR, so maybe it might bring back the several volumes per disk crash https://github.com/pantavisor/pantavisor/pull/205

PR #210, opened 2023-01-24
fix: shell activation depending on ssh config key

anibalportero:


init.c
```diff
@@ -531,13 +531,13 @@ int main(int argc, char *argv[])
 		goto loop;
 
 	// these debugging tools will be children of the pv thread, so we can controll them
-	if (pv_config_get_debug_ssh()) {
-		if (pv_config_get_debug_shell())
-			debug_shell();
+	if (pv_config_get_debug_ssh())
 		debug_telnet();
-	} else {
+	else
```
asac, line 536: please no 'else' without brackages if the if has themm...

anibalportero: its an else to pv_config_get_debug_ssh, indentation was bad

PR #212, opened 2023-01-25
fix: disk transfer in non-reboot updates not taking place

anibalportero:
When we do a non-reboot updates, we transfer objects, jsons, platforms and volumes from the new state struct (that will be later discarded) to the current one (the one that was booted, which will be preserved after non-reboot transition). We also have to relink the new volumes from the new state to the disks in the old state, otherwise, those will be pointing disks from the new state that will be freed.

### DIFF
--- a/state.c
+++ b/state.c
@@ -216,6 +216,24 @@ struct pv_json *pv_state_fetch_json(struct pv_state *s, const char *name)
 	return NULL;
 }
 
+static struct pv_disk *pv_state_fetch_disk(struct pv_state *s, const char *name)
+{
+	struct pv_disk *d, *tmp;
+
+	if (!name)
+		return NULL;
+
+	// Iterate over all disks from state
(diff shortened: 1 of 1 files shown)


state.c
```diff
@@ -961,6 +980,13 @@ static void pv_state_transfer_platforms(struct pv_state *pending,
 		       v->name, v->plat->name);
 		dl_list_del(&v->list);
 		dl_list_add_tail(&current->volumes, &v->list);
+
+		if (v->disk) {
+			d = pv_state_fetch_disk(current, v->disk->name);
+			pv_log(DEBUG, "relinking volume %s to disk %s", v->name,
+			       d->name);
+			v->disk = d;
```
asac, line 988: is this helping without making a copy of the disk from old state or am i missing something?

anibalportero: the disks are already there. The current one is the one that we have used in the old revision, and the one that we are going to use the new one too. We are just linking new volumes to the old revision disk here, as the one from the new state will be removed.

PR #215, opened 2023-01-27
fix: gc checking if pantahub.config exists

anibalportero:
This is subsituted by creating a file /storage/config/mounted after mounting /storage. That file is then checked before getting storage info for GC and logs.

### DIFF
--- a/paths.h
+++ b/paths.h
@@ -64,6 +64,7 @@ void pv_paths_storage_devmeta(char *buf, size_t size);
 void pv_paths_storage_dropbear(char *buf, size_t size);
 
 #define COREPV_FNAME "corepv"
+#define PVMOUNTED_FNAME ".pvmounted"
 
 void pv_paths_storage_file(char *buf, size_t size, const char *name);
 void pv_paths_storage_object(char *buf, size_t size, const char *sha);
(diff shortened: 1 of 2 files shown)


paths.h
```diff
@@ -87,6 +87,7 @@ void pv_paths_storage_trail_pvr_file(char *buf, size_t size, const char *rev,
 
 #define UNCLAIMED_FNAME "unclaimed.config"
 #define PANTAHUB_FNAME "pantahub.config"
+#define MOUNTED_FNAME "mounted"
```
asac, line 90: lets name it .pv_mounted .... same we probably would want to add to all disks we mount (e.g. crypt disks)... and later then also gather space of those disks of course...

asac (changes requested):

paths.h
anibalportero: done

storage.c
```diff
@@ -220,7 +220,7 @@ static struct pv_storage *pv_storage_new()
 	struct statfs buf;
 	struct pv_storage *this;
 
-	pv_paths_storage_config_file(path, PATH_MAX, PANTAHUB_FNAME);
+	pv_paths_storage_file(path, PATH_MAX, PVMOUNTED_FNAME);
```
asac, line 223: why this change?

anibalportero: well, I imagined if we are going to use .pvmounted for disks too why not storing that in /storage instead of /storage/config

asac (approved):

PR #227, opened 2023-02-08
allow to explicitely select user-meta location as authorized_keys 

asac:
…h pv:user-meta as config value

### DIFF
--- a/debug.c
+++ b/debug.c
@@ -94,7 +94,8 @@ void pv_debug_start_ssh()
 	if (db_pid > -1)
 		return;
 
-	if (!pv_config_get_debug_ssh_authorized_keys())
+	if (!pv_config_get_debug_ssh_authorized_keys() ||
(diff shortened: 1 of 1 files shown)


debug.c
```diff
@@ -94,7 +94,8 @@ void pv_debug_start_ssh()
 	if (db_pid > -1)
 		return;
 
-	if (!pv_config_get_debug_ssh_authorized_keys())
+	if (!pv_config_get_debug_ssh_authorized_keys() ||
+	    !strcmp(pv_config_get_debug_ssh_authorized_keys(), "pv:user-meta"))
```
anibalportero, line 98: why not just using "default" instead? that way it makes more sense at least from the point of view of pvcontrol

pvcontrol usrmeta save debug.ssh_authorized_keys default

asac: ok using __default__ now

PR #243, opened 2023-03-24
add pvcrash handler and stop forcefully registering core_pattern the old way

asac:
You can now use pantavisor.config sysctl.kernel.core_pattern=|/lib/pv/pvcrash /storage/crash/ to enable crash dumps to /storage/crash/

### DIFF
--- a/pantavisor.c
+++ b/pantavisor.c
@@ -953,7 +953,6 @@ static pv_state_t _pv_run_state(pv_state_t state, struct pantavisor *pv)
 
 int pv_start()
 {
-	char path[PATH_MAX];
 	struct pantavisor *pv = pv_get_instance();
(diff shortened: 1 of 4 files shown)


pantavisor.c
```diff
@@ -974,14 +973,6 @@ int pv_start()
 
 	setrlimit(RLIMIT_CORE, &core_limit);
 
-	pv_paths_storage_file(path, PATH_MAX, COREPV_FNAME);
```
anibalportero, line 977: so where are they going to be stored now with this patch?

asac: whatever is in sysctl.kernel.core_pattern in pantavisor.config ... in the last commit i added a default maybe see what i did.

PR #246, opened 2023-03-24
unify approach of remount.sh and support profiles (default is default)

asac:
Can be configured system global defaults in in device.json/remount and refined in run.json/remount


asac:
still broken.

PR #250, opened 2023-03-29
introduce group and state status for logging

anibalportero:
This PR introduces status for groups. This allows us to monitor and log status changes, so we know when a group is STARTED or READY, for example.

The status of a group is always READY, except when any of the platforms that form the group have not achieved their status goals. In that case, the status of the group corresponds with the platform with the lower status.

It also adds status of the state revision. This status is calculated the same way as the group status, but having into account all platforms in the revision instead.

### DIFF
--- a/platforms.c
+++ b/platforms.c
@@ -130,12 +130,41 @@ const char *pv_platform_status_string(plat_status_t status)
 	return "UNKNOWN";
 }
 
+static void pv_platform_log_timer_status(struct pv_platform *p)
+{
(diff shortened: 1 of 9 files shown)


platforms.c
```diff
@@ -136,6 +136,10 @@ static void pv_platform_set_status(struct pv_platform *p, plat_status_t status)
 		return;
 
 	p->status.current = status;
+	pv_log(DEBUG, "platform '%s' status is now %s", p->name,
```
asac, line 139: if this is the real status message that we are after, lets use INFO level at least.....

PR #274, opened 2023-04-10
Remove useless assignments from several files

fortizc:
Remove some assignments detected as "never read"
Files:
* ctrl.c
* drivers.c
* grub.c
* logserver
* loop.c
* metadata.c
* network.c
* parser/parser_multi1.c
* parser/parser_system1.c
* ph_logger.c
* signature.c
* storage.c
* trestclient.c
* uboot.c
* updater.c
* utils/tsh.c
* wdt.c

### DIFF
--- a/drivers.c
+++ b/drivers.c
@@ -42,7 +42,7 @@
 static char *_sub_meta_values(char *str)
 {
 	// find key for meta from ${user-meta:KEY} component of string
-	int _start = 0, _end = 0, type;
+	int _start = 0, _end = 0;
 	char *_tok, *_t, *_at, *new;
 	char *_var, *_param, *_tstr;
 	int _nt = 0;
@@ -65,13 +65,8 @@ static char *_sub_meta_values(char *str)
 		_nt = _nt + (_at - (str + _end));
 		_var = strchr(_t, '}');
 		_end = _var - (str);
-		_start = _at - (str + _start);
 		_tstr = strdup(_t);
-		_tok = strtok(_tstr, "{:}");
-		if (!strcmp(_tok, USER_META_KEY))
-			type = USER_META;
-		else if (!strcmp(_tok, DEV_META_KEY))
-			type = DEVICE_META;
+		strtok(_tstr, "{:}");
 		_tok = strtok(NULL, "{:}");
 		_param = pv_metadata_get_usermeta(_tok);
 		if (_param) {
--- a/ph_logger.c
+++ b/ph_logger.c
@@ -356,7 +356,6 @@ static int ph_logger_push_from_file(const char *filename, char *platform,
 				 * The reason being we can't differentiate between a slow
 				 * growing file and a file that doesn't grow at all.
 				 */
-				bytes_read = 0;
 				break;
 			}
 		}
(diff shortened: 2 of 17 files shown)


wdt.c
```diff
@@ -37,8 +37,6 @@ int fd = -1;
 
 int pv_wdt_start(struct pantavisor *pv)
 {
-	int timeout = pv_config_get_watchdog_timeout();
-	int interval = pv_config_get_updater_interval();
```
fortizc, line 41: I have some doubts with this variable, if you see in the original code, the value from `pv_config_get_updater_interval()` was never used, but it could be, right? the solution only mimics the previous behavior, but is (and was) a bug, or not?

ph_logger.c
```diff
@@ -356,7 +356,6 @@ static int ph_logger_push_from_file(const char *filename, char *platform,
 				 * The reason being we can't differentiate between a slow
 				 * growing file and a file that doesn't grow at all.
 				 */
-				bytes_read = 0;
```
anibalportero, line 359: unsure about this one too... It's inside a while and we check bytes_read in its condition

wdt.c
anibalportero: I guess we put that on mistake in there after a config refactor. Previous to that, the config updater.interval was overridden based on the timeout value, which is not documented and a hidden weird behavior. I'd remove the "int interval = timeout / 2;" line, which is not doing anything, and the "pv_log(INFO..." that follows

drivers.c
```diff
@@ -65,13 +65,8 @@ static char *_sub_meta_values(char *str)
 		_nt = _nt + (_at - (str + _end));
 		_var = strchr(_t, '}');
 		_end = _var - (str);
-		_start = _at - (str + _start);
```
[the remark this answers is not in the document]
fortizc: Yes, is used, but always it will be replaced with the value of `_end + 1` at the end of the loop and in between of those two assignments is never used

drivers.c
```diff
@@ -65,13 +65,8 @@ static char *_sub_meta_values(char *str)
...
 		_tstr = strdup(_t);
-		_tok = strtok(_tstr, "{:}");
```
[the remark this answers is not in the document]
fortizc: clang-tidy mark that as warning because the value of `_tok` is never read, it replaced by the next `strtok` call there are 2 consecutive calls as I remove the `if-else if` block

ph_logger.c
fortizc: Yes, but the next instruction is a `break` and `bytes_read` isn't check after the loop

drivers.c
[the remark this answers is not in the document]
anibalportero: true

drivers.c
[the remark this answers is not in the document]
anibalportero: you are right

ph_logger.c
anibalportero: indeed

wdt.c
fortizc: So, we remove all traces of `interval` in that function?

fortizc: ok

PR #279, opened 2023-04-14
Removes all memory leaks founds by clang-tidy

fortizc:


parser/parser_system1.c
```diff
@@ -174,16 +174,18 @@ static int parse_bsp_drivers(struct pv_state *s, char *v, int len)
...
 		return 0;
 
+	buf = memcpy(buf, v, len);
+
 	jsmnutil_parse_json(buf, &tokv, &tokc);
 
 	keys = jsmnutil_get_object_keys(buf, tokv);
 	if (!keys) {
 		pv_log(ERROR, "drivers list cannot be parsed");
+		free(buf);
```
anibalportero, line 188: github doesn't let me coment in line 207, but we would miss a "free(buf);" before "return 0;" I suppose

PR #285, opened 2023-04-18
Fix clang-tidy warnings

fortizc:
Different types of problems are included, problems with memory already freed, uninitialized variables, etc

### DIFF
--- a/buffer.c
+++ b/buffer.c
@@ -105,6 +105,8 @@ static int pv_buffer_init_cache(int items, int size, struct dl_list *head)
 			free(item);
 		}
 	}
+
+	dl_list_init(head);
 	while (items > 0) {
 		struct buffer *buffer = NULL;
 
--- a/ctrl.c
+++ b/ctrl.c
@@ -602,11 +602,14 @@ out:
 		close(obj_fd);
 }
 
-static char *pv_ctrl_get_file_name(const char *path, int buf_index,
+static char *pv_ctrl_get_file_name(const char *path, size_t buf_index,
 				   size_t path_len)
 {
 	int len;
 	char *file_name;
+
+	if (buf_index > path_len)
+		return NULL;
 
 	len = path_len - buf_index;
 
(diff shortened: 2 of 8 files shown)


log.h
```diff
@@ -30,7 +30,7 @@
 #include "pantavisor.h"
 #include <stdarg.h>
 
-void exit_error(int err, char *msg);
+void exit_error(int err, char *msg) __attribute__((__noreturn__));
```
fortizc, line 33: FYI: this it seems to be mandatory for this kind of function on the clang-tidy are called [custom assertion handlers](https://clang-analyzer.llvm.org/annotations.html#custom_assertions)

buffer.c
```diff
@@ -105,6 +105,8 @@ static int pv_buffer_init_cache(int items, int size, struct dl_list *head)
 			free(item);
 		}
 	}
+
+	dl_list_init(head);
```
anibalportero, line 109: It amazes me that this wasn't initialized all this time... The variable head is still used previously in this function though. Don't get me wrong, I think this is the place for this dl_list_init, but the previous code should be moved to pv_buffer_free or similar and I don't understand what was this doing here.

Seeing buffer in a general way, as it is used by metadata, ph_logger and logserver, maybe it shouldn't be initialized from log.c either and moved it to a more general location. Besides this, the new pv_buffer_free would have to be called somewhere at the end of pv_shutdown so we can fix this mem leak.

If you think it's a lot for this PR, we can leave it for a further one.

ctrl.c
```diff
@@ -610,6 +610,9 @@ static char *pv_ctrl_get_file_name(const char *path, int buf_index,
 
 	len = path_len - buf_index;
 
+	if (len < 0)
```
anibalportero, line 613: IMO len and buf_index should be size_t and just check at the beginning if buf_index > path_len

log.h
anibalportero: I don't understand this very well. Does it mean all void returning functions should have this?

fortizc: Not all, only those finish the execution. It seems clang-tidy sometimes has problems to recognize when a function is a "custom assertion handler"

ctrl.c
fortizc: Right, I will change it

buffer.c
fortizc: Yeah, probably we need a new PR to fix that maybe pv_buffer_init() and pv_buffer_free() I think that both must be called by any file or library that needs a buffer and internally the buffer could work with reference counting, so if it's already initialized pv_buffer_init() does nothing except add 1 to the ref counter. If pv_buffer_free() is called decrease the counter, if the counter is equals to 0 run the routine to free all the resources.

PR #288, opened 2023-04-21
Remove strcat and strcpy from pantavisor

fortizc:


updater.c
```diff
@@ -125,7 +127,7 @@ static char *unescape_utf8_to_apvii(char *buf, char *code, char c)
 		p = strstr(tmp, code);
 	}
 
-	if (new && new[strlen(new) - 1] == c)
+	if (new &&new[strlen(new) - 1] == c)
```
anibalportero, line 130: is this one on purpose?

PR #290, opened 2023-04-25
fix: group order not being honored after timeout

anibalportero:
This PR fixes the platform starting order after any platform times out because of failing status goal. After this fix, it follows the group order instead of the alphabetical order of the platforms.

The problem was happening because pv_state_check_goals was overwriting whatever value was in the status goal of the immediately previous group if any of the previous groups behind have had a timeout.

For that, the old pv_state_check_goals function has been divided to make the code more clear. Instead of having the same function performing two different things depening on p being NULL or not. The old pv_state_check_goals now returns the state of the goal of the whole revision only, while the new pv_state_check_goal_prev_group returns the state of the goals of the platforms belonging to the previous group.


state.c
```diff
@@ -515,6 +512,26 @@ plat_goal_state_t pv_state_check_goals(struct pv_state *s,
...
+
+	plat_goal_state_t g_goal_state = PLAT_GOAL_ACHIEVED;
+
+	struct pv_group *g, *tmp;
+	dl_list_for_each_safe(g, tmp, &s->groups, struct pv_group, list)
+	{
+		if (group == g)
+			break;
+
+		g_goal_state = pv_group_check_goals(g);
```
fortizc, line 529: Here you don't need to do something like this?
```
if (g_goal_state > s_goal_state)
			s_goal_state = g_goal_state;
```

anibalportero: that was actually how it was before. In the case of pv_state_check_goals that's OK, as any timeout in the state revision means we have to rollback but, in the case of pv_state_check_goal_prev_group, its not OK, as any timeout in the sate revision will override the return value of the function and ignore the value that we want, which is just the one from the previous group of the group we are evaluating

PR #300, opened 2023-05-26
Fixes stdout log messages when logserver is not running

fortizc:
Adds function to send logs formatted to stdout when logserver is not running

### DIFF
--- a/logserver/logserver.c
+++ b/logserver/logserver.c
@@ -166,7 +166,7 @@ static int pv_log(int level, char *msg, ...)
 			return 1;
 
 		vsnprintf(log.data.buf, log.data.len, msg, args);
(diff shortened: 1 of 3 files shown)


logserver/logserver.c
```diff
@@ -950,40 +944,44 @@ int pv_logserver_send_vlog(bool is_platform, char *platform, char *src,
 		.src = src,
 	};
 
-	if (logserver.pid <= 0 || level > pv_config_get_log_loglevel())
+	if (level > pv_config_get_log_loglevel())
 		return -1;
```
anibalportero, line 948: shouldn't this return 0? Doesn't look like an error if the level of the message doesn't make it because of loglevel

fortizc: Yes, you're right!

PR #314, opened 2023-06-09
fix: empty reply from pv-ctrl

anibalportero:


ctrl.c
```diff
@@ -85,12 +85,12 @@
 #define HTTP_RES_CONT "HTTP/1.1 100 Continue\r\n\r\n"
 
 #define HTTP_RESPONSE                                                          \
-	"HTTP/1.1 %s \r\nContent-Length: %d\r\nContent-Type: application/json; charset=utf-8\r\n\r\n{\"Error\":\"%s\"}\r\n"
+	"HTTP/1.1 %s \r\nContent-Length: %zd\r\nContent-Type: application/json; charset=utf-8\r\n\r\n{\"Error\":\"%s\"}\r\n"
```
asac, line 88: can we elaborate in the commit message a bit more what we saw and why and what we change to prevent that now and why?

PR #323, opened 2023-06-14
fix: pv-ctrl empty reply 

anibalportero:
This was found using the pvcontrol steps put with a bad name, which in some cases was returning:

```
*   Trying /var/run/pv-ctrl.pantavisor:0...
* Connected to localhost (/var/run/pv-ctrl.pantavisor) port 80 (#0)
> PUT /steps/test HTTP/1.1
> Host: localhost
> User-Agent: curl/8.1.2
> Accept: */*
> Content-Length: 4588
>
} [4588 bytes data]
* We are completely uploaded and fine
* Empty reply from server
* Closing connection 0
```

This has been fixed by flushing every pending to read byte in case we have to return an error in PUT or POST before reading the request body.

### DIFF
--- a/ctrl.c
+++ b/ctrl.c
@@ -216,6 +216,20 @@ out:
 	return res;
 }
 
+static void pv_ctrl_flush_req(int req_fd, size_t content_length)
+{
+	char buf[HTTP_REQ_BUFFER_SIZE];
+	size_t total_rec = 0;
+	ssize_t rec;
+
+	while ((rec = read(req_fd, buf, content_length - total_rec)) > 0) {
+		if (rec < 0)
+			return;
+
+		total_rec += rec;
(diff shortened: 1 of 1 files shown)


ctrl.c
```diff
@@ -227,6 +241,7 @@ static int pv_ctrl_process_signal(int req_fd, size_t content_length,
 	pv_log(DEBUG, "reading and parsing signal...");
 
 	if (content_length >= HTTP_REQ_BUFFER_SIZE) {
+		pv_ctrl_flush_req(req_fd, content_length);
```
asac, line 244: i find the name of the function confusing as it is not flushing, but rather consuming ... so maybe lets rename it so we dont get confused in future.

PR #324, opened 2023-06-14
Add system.mdev option to load drivers using mdev

fortizc:
- Adds interface to load/unload a single driver
- Adds support to load drivers on early stage for standalone and embedded modes
- Cleans initial modules and fw

### DIFF
--- a/config.c
+++ b/config.c
@@ -350,6 +350,8 @@ static int pv_config_load_file(char *path, struct pantavisor_config *config)
 		&config_list, "system.mediadir", "/media");
 	config->sys.confdir = config_get_value_string(
 		&config_list, "system.confdir", "/configs");
+	config->sys.auto_load_drivers = config_get_value_bool(
+		&config_list, "system.drivers.load_early.auto", false);
 
 	config->policy = config_get_value_policy(&config_list, "policy", NULL);
 	if (pv_config_load_policy(config->policy, &config_list))
@@ -933,6 +935,11 @@ char *pv_config_get_system_confdir()
 	return pv_get_instance()->config.sys.confdir;
 }
 
+bool pv_config_get_system_early_driver_load()
+{
+	return pv_get_instance()->config.sys.auto_load_drivers;
+}
+
 bool pv_config_get_debug_shell()
 {
 	return pv_get_instance()->config.debug.shell;
@@ -1279,6 +1286,8 @@ char *pv_config_get_json()
 		pv_json_ser_string(&js, pv_config_get_system_mediadir());
 		pv_json_ser_key(&js, "system.confdir");
 		pv_json_ser_string(&js, pv_config_get_system_confdir());
+		pv_json_ser_key(&js, "system.drivers.load_early.auto");
+		pv_json_ser_bool(&js, pv_config_get_system_early_driver_load());
 		pv_json_ser_key(&js, "debug.shell");
 		pv_json_ser_bool(&js, pv_config_get_debug_shell());
 		pv_json_ser_key(&js, "debug.shell.autologin");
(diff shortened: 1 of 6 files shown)


init.c
```diff
@@ -387,6 +409,8 @@ int main(int argc, char *argv[])
 		goto loop;
 	}
 
+	load_early_drivers(pv_config_get_system_mdev());
```
anibalportero, line 412: I this as late in the init process as it can be? I guess it cannot be added to pv_init_tbl? In that case, do we want this to be executed in IM_STANDALONE and IM_APPENGINE? If so, maybe we should control what we are loading to then be able to unload it during shutdown. Either way, remember to test this in appengine when you get it ready :)

init.c
```diff
@@ -68,6 +69,27 @@
 #define MAX_PROC_STATUS (10)
 pid_t pv_pid;
 
+static void load_early_drivers(bool run_mdev)
```
anibalportero, line 72: Wouldn't this be better in drivers.c? Also, woudln't it be better to use pv_config_get_system_mdev inside instead of using an argument?

drivers.c
```diff
@@ -135,6 +138,44 @@ static int _pv_drivers_modprobe(char **modules, mod_action_t action)
...
+	char *cmd = calloc(len, sizeof(char));
+	if (!cmd)
+		return -1;
+
+	snprintf(cmd, len, DRIVER_MGMT_COMMAND, name, ld, op);
+	int status = 0;
+	tsh_run(cmd, 0, &status);
+	status = WEXITSTATUS(status);
+	if (status != 0) {
+		pv_log(WARN, "cannot %s module %s with oprions: %s", name,
```
anibalportero, line 161: typo

fortizc: :P

init.c
fortizc: > I this as late in the init process as it can be? I guess it cannot be added to pv_init_tbl?

I guess this should be before to call `other_mounts()` because we need the fs driver and probably the `usb-storage`. 

> do we want this to be executed in IM_STANDALONE and IM_APPENGINE?

could be... but that'd be another target? something like `x64-ubuntu-lts-appengine` ??

anibalportero: I think there is no need to add is as a target for now. I was more interested in what new stuff is going to be executed here for the existing appengine targets  (and for other targets now that I think about it). I guess, according to the "system.mdev" parameter, we are going to still do "pv_drivers_load_single("usb-storage", NULL);" for the targets that have it disabled. Is this what we want?

drivers.c
```diff
@@ -39,6 +40,9 @@
 #define USER_META_KEY "user-meta"
 #define DEV_META_KEY "device-meta"
 
+// modprobe + remove option + module name + module options
+#define DRIVER_MGMT_COMMAND "/sbin/modprobe %s %s %s"
```
asac, line 44: is it possible to call kmod directly? In that way its easier to understand why something isn't working if you havent enabled KMOD in global.config....

config.c
```diff
@@ -326,6 +326,8 @@ static int pv_config_load_file(char *path, struct pantavisor_config *config)
 		&config_list, "system.mediadir", "/media");
 	config->sys.confdir = config_get_value_string(
 		&config_list, "system.confdir", "/configs");
+	config->sys.mdev =
+		config_get_value_bool(&config_list, "system.mdev", false);
```
anibalportero, line 330: sorry for being so annoying with the names. Can we rename this to system.driver.mdev or system.drivers.mdev?

drivers.c
fortizc: Ok, got it, but my initial idea was to replace the old modprobe invocation in https://github.com/pantavisor/pantavisor/blob/feature/mdev/drivers.c#L164 with this in the future. I don't really like to have 2 places where we use a very similar command. So the idea was to use `pv_drivers_load_single` and `pv_drivers_unload_single`.

drivers.c
```diff
@@ -135,6 +177,25 @@ static int _pv_drivers_modprobe(char **modules, mod_action_t action)
 	return ret;
 }
 
+void pv_drivers_load_early()
+{
+	if (!pv_config_get_system_mdev())
```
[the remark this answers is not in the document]
fortizc: Reading this comment and the comment from @anibalportero I think that maybe is better `system.drivers.load_early.auto=1` ?

fortizc:
@asac @anibalportero I changed both the option name and the function name in config.c please check it

config.c
asac: yes lets rename

PR #333, opened 2023-07-04
filter lxc cgroup v1 config items in v2 and viceversa

anibalportero:
filter out all v2 cgroup allow and deny keys if in v1. Filter out all v1 keys if in v2. Keep everything if in hybrid

### DIFF
--- a/plugins/pv_lxc.c
+++ b/plugins/pv_lxc.c
@@ -207,26 +207,26 @@ static int pv_setup_config_bindmounts(struct lxc_container *c, char *srcdir,
 	return 1;
 }
 
-static void pv_setup_lxc_container_cgroup(struct lxc_container *c)
+static void pv_setup_lxc_filter_cgroup(struct lxc_container *c)
 {
-	// only for cgroup unified
-	if (__pv_get_instance()->cgroupv != CGROUP_UNIFIED)
-		return;
-
-	// XXX: this might change with LXC 5.0, as we might be able to support the original lxc.conf
-
 	char value[PATH_MAX];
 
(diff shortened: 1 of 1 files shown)


plugins/pv_lxc.c
```diff
@@ -207,26 +207,26 @@ static int pv_setup_config_bindmounts(struct lxc_container *c, char *srcdir,
...
-
 	char value[PATH_MAX];
 
-	// remove all ilegacy cgroup allow and deny config
-	while (c->get_config_item(c, "lxc.cgroup.devices.allow", value, PATH_MAX) > 0) {
-		c->set_config_item(c, "lxc.cgroup.devices.allow", NULL);
-	}
-	while (c->get_config_item(c, "lxc.cgroup.devices.deny", value, PATH_MAX) > 0) {
-		c->set_config_item(c, "lxc.cgroup.devices.deny", NULL);
+	// filter all v2 config items while in v1 and viceversa
```
asac, line 214: I think we should only do the removal if we founud that both lxc.cgroup as well as lxc.cgroup2 are in the config (interpreting this as an intend that both worlds have been modelled) ... similarly if we detect that are having just the wrong cgroups we should error.

anibalportero: this would mean that we have to add this double cgroup config to all containers with pvr, right?

asac: yes, this means in order for containers to support both cgroup worlds, they have to come with info about how to deal with both cgroup worlds. Similarly, I think we can also admit a container for both worlds if the container has ZERO lxc.cgroup or lxc2.cgroup configs...

PR #334, opened 2023-07-13
add wdt startup and always operation modes

anibalportero:
This PR adds a new config key named "wdt.mode" with four possible options:
* disabled: same as old wdt.enabled=0
* shutdown (default): same as old wdt.enabled=1
* startup: the watchdog will be pinged from initialization till the first container is started
* always: the watchdog will be pinged during all Pantavisor operation

This relies on this old kick call that is performed before progressing to any state:

```
static pv_state_t _pv_run_state(pv_state_t state, struct pantavisor *pv)                                              
{
    pv_wdt_kick(pv);
    return state_table[state](pv);                                                                                    
}
```

And assumes that wdt.timeout will be configured with a value that is bigger than any possible state transition.

### DIFF
--- a/wdt.c
+++ b/wdt.c
@@ -33,40 +33,50 @@
 #define pv_log(level, msg, ...) vlog(MODULE_NAME, level, msg, ##__VA_ARGS__)
 #include "log.h"
 
-int fd = -1;
+static int pv_wdt_fd = -1;
 
 int pv_wdt_start(struct pantavisor *pv)
 {
-	if (!pv_config_get_watchdog_enabled())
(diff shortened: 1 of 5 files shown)


wdt.c
```diff
@@ -58,14 +59,23 @@ int pv_wdt_start(struct pantavisor *pv)
...
+		return;
+
+	close(fd);
+	fd = -1;
+}
+
 void pv_wdt_kick(struct pantavisor *pv)
 {
-	if (!fd)
+	if (fd < 0)
```
asac, line 78: nice!

PR #341, opened 2023-08-07
Mounts /sys/kernel/security

fortizc:
Mounts `/sys/kernel/security` needed by AppArmor

### DIFF
--- a/config.c
+++ b/config.c
@@ -352,6 +352,8 @@ static int pv_config_load_file(char *path, struct pantavisor_config *config)
 		&config_list, "system.confdir", "/configs");
 	config->sys.auto_load_drivers = config_get_value_bool(
 		&config_list, "system.drivers.load_early.auto", false);
+	config->sys.mount_securityfs = config_get_value_bool(
+		&config_list, "system.mount.securityfs", false);
 
 	config->policy = config_get_value_policy(&config_list, "policy", NULL);
 	if (pv_config_load_policy(config->policy, &config_list))
@@ -940,6 +942,11 @@ bool pv_config_get_system_early_driver_load()
 	return pv_get_instance()->config.sys.auto_load_drivers;
 }
 
+bool pv_config_get_system_mount_securityfs()
+{
+	return pv_get_instance()->config.sys.mount_securityfs;
+}
+
 bool pv_config_get_debug_shell()
 {
 	return pv_get_instance()->config.debug.shell;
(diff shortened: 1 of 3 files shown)


init.c
```diff
@@ -87,6 +88,10 @@ static int early_mounts()
 	if (ret < 0)
 		exit_error(errno, "Could not mount /sys");
 
+	ret = mount("none", "/sys/kernel/security", "securityfs", 0, NULL);
```
asac, line 91: some systems might not have securityfs support built into kernel ... so have this a config that is default on?

fortizc: Ok, got it, I gonna add a config to this, something like `system.mount.sys.kernel.security` as a boolean and move this code to `other_mounts()` in order to get the config system ready

fortizc: New key added to config

config.c
```diff
@@ -352,6 +352,8 @@ static int pv_config_load_file(char *path, struct pantavisor_config *config)
 		&config_list, "system.confdir", "/configs");
 	config->sys.auto_load_drivers = config_get_value_bool(
 		&config_list, "system.drivers.load_early.auto", false);
+	config->sys.mount_kernel_sec = config_get_value_bool(
+		&config_list, "system.mount.kernel_sec", false);
```
anibalportero, line 356: I propose system.mount.securityfs

fortizc:
The configurations were renamed. Please check the new name.

config.c
```diff
@@ -1288,6 +1295,8 @@ char *pv_config_get_json()
 		pv_json_ser_string(&js, pv_config_get_system_confdir());
 		pv_json_ser_key(&js, "system.drivers.load_early.auto");
 		pv_json_ser_bool(&js, pv_config_get_system_early_driver_load());
+		pv_json_ser_key(&js, "system.mount_securityfs");
```
anibalportero, line 1298: you forgot to change it here :)

config.c
```diff
@@ -1459,6 +1468,8 @@ void pv_config_print()
 	pv_log(INFO, "system.configdir = '%s'", pv_config_get_system_confdir());
 	pv_log(INFO, "system.drivers.load_early.auto = %d",
 	       pv_config_get_system_early_driver_load());
+	pv_log(INFO, "system.mount_securityfs = %d",
```
anibalportero, line 1471: also here

PR #342, opened 2023-08-16
Adds config key to load lxc apparmor profiles

fortizc:
Adds `system.load_lxc_profiles` to load lxc profiles for apparmor

### DIFF
--- a/atom.mk
+++ b/atom.mk
@@ -97,7 +97,8 @@ LOCAL_SRC_FILES := debug.c \
 			blkid.c \
 			paths.c \
 			drivers.c \
-			cgroup.c
+			cgroup.c \
+			apparmor.c
 
 LOCAL_INSTALL_HEADERS := log.h
 LOCAL_GENERATED_SRC_FILES := version.c
--- a/init.c
+++ b/init.c
@@ -52,6 +52,7 @@
 #include "cgroup.h"
 #include "wdt.h"
 #include "drivers.h"
+#include "apparmor.h"
 
 #include "utils/tsh.h"
 #include "utils/math.h"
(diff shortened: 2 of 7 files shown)


atom.mk
```diff
@@ -97,7 +97,8 @@ LOCAL_SRC_FILES := debug.c \
 			blkid.c \
 			paths.c \
 			drivers.c \
-			cgroup.c
+			cgroup.c \
+			apparmor_loader.c
```
anibalportero, line 101: I propose apparmor.c to keep consistency with the rest of modules

apparmor_loader.h
```diff
@@ -0,0 +1,28 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#ifndef PV_APPARMOR_LOADER_H
+#define PV_APPARMOR_LOADER_H
+
+void load_apparmor_profiles(void);
```
anibalportero, line 26: The convention is to use the name of the module as a prefix like: pv_apparmor_load_profiles

apparmor_loader.c
```diff
@@ -0,0 +1,105 @@
...
+
+#include "apparmor_loader.h"
+#include "utils/tsh.h"
+#include "config.h"
+
+#define MODULE_NAME "apparmor_loader"
+#define pv_log(level, msg, ...) vlog(MODULE_NAME, level, msg, ##__VA_ARGS__)
+#include "log.h"
+
+#define PV_APPARMOR_PARSER_CMD "apparmor_parser -r %s/%s"
```
anibalportero, line 40: where is apparmor_parser installed? asking in case this can be made portable for appengine

apparmor_loader.c
```diff
@@ -0,0 +1,105 @@
...
+#define PV_APPARMOR_PROFILES_DIR "/etc/apparmor.d"
```
anibalportero, line 41: shouldn't be this according to system.etcdir instead of just "/etc"?

init.c
```diff
@@ -408,6 +409,8 @@ int main(int argc, char *argv[])
 		exit_error(errno, "Could not init cgroup");
 	other_mounts();
 
+	load_apparmor_profiles();
```
anibalportero, line 412: why here? just asking so I get who is using this etc. :)

load_apparmor_profiles would be executed in appengine mode, so it would be nice to have the aforementioned configurable path stuff

apparmor_loader.c
fortizc: Ok, I'm going to add the complete path

init.c
anibalportero: That's OK. I was meaning more if we are calling load_apparmor_profiles as late as possible (Notice that we don't have logserver initialized and that's inconvinient for debugging, for example)

apparmor_loader.c
anibalportero: It depends, if the thing is installed by us, in order to work with appengine, it is necessary to use a configurable path like we would do with system.etcdir

anibalportero:
looks good to me now :)

PR #344, opened 2023-08-29
Improve WONTGO status message

anibalportero:
This is the first delivery in the [improvements of update error reports](https://hackmd.io/33RAIpSSS6SOGj2lDTyAfQ).

Here, we are expand the information given in status-msg in the progress JSON, which can be accessed by `pvcontrol steps show-progress` and `pvcontrol post` commands, following this table:

| Before | After |
| ------ | ----- |
State signatures cannot be verified | Secureboot: State not fully covered by signatures |
State signatures cannot be verified | Secureboot: Signature validation failed |
State cannot be parsed | Parser: State JSON has bad format |
Update aborted | PH Client: Update aborted |
Unable to download and/or install update | PH Client: Max download retries reached |
Space required X B, available Y B | PH Client: Space required X B, available Y B |

### DIFF
--- a/signature.c
+++ b/signature.c
@@ -911,9 +911,10 @@ out:
 	return ret;
 }
 
-static bool pv_signature_verify_pairs(struct dl_list *json_pairs)
+static sign_state_res_t pv_signature_verify_pairs(struct dl_list *json_pairs)
 {
-	bool ret = true, found = false;
+	bool found = false;
+	sign_state_res_t ret = SIGN_STATE_OK;
 	struct pv_signature_pair *pair, *tmp;
 	struct pv_signature *signature = NULL;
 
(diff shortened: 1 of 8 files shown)


signature.c
```diff
@@ -688,12 +690,12 @@ static int pv_signature_validate_pk(struct mbedtls_pk_context *pk)
...
-				    struct dl_list *certs_raw,
-				    struct pv_signature *signature,
-				    mbedtls_md_type_t mdtype)
+static sign_file_res_t pv_signature_verify_sha(const char *payload,
+					       struct dl_list *certs_raw,
+					       struct pv_signature *signature,
+					       mbedtls_md_type_t mdtype)
 {
-	bool ret = false;
+	bool ret = SIGN_FILE_NOK_SHA;
```
fortizc, line 698: This is weird, this bool should be `sign_file_res_t` ?

anibalportero: that's bad. Good catch :)

PR #346, opened 2023-09-13
Adds timestamp in singlefile json log

fortizc:
- Adds `log.timestamp.format`
- Adds some shorthand for well known timestamp format (WIP)
- Change how the JSON log is built, now use the PV JSON lib

### DIFF
--- a/atom.mk
+++ b/atom.mk
@@ -48,6 +48,7 @@ LOCAL_SRC_FILES := debug.c \
 			loop.c \
 			logger.c \
 			log.c \
+			logserver/logserver_timestamp.c \
 			logserver/logserver_utils.c \
 			logserver/logserver_out.c \
 			logserver/logserver_singlefile.c \
--- a/config.c
+++ b/config.c
@@ -468,6 +468,12 @@ static int pv_config_load_file(char *path, struct pantavisor_config *config)
 				     &config->log.logdir);
 	config->log.dmesg =
 		config_get_value_bool(&config_list, "log.capture.dmesg", false);
+	config->log.ts_filetree_fmt = config_get_value_string(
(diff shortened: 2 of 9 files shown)


anibalportero:
Is this applicable to filetree/stdout? one would imagine a key named log.timestamp.format would not discern between logserver outputs.

Also, remember to add the new key to https://docs.pantahub.com/pantavisor-configuration/#pantavisorconfig once you are finished :)

fortizc:
> Is this applicable to filetree/stdout? one would imagine a key named log.timestamp.format would not discern between logserver outputs.

Yes, you're definitely right, is it a good idea to add the TS to all the outputs?
currently, the format is as follows:
```
[pantavisor] 11 INFO     -- [config]: system.rundir = '/pv'
```
adding TS info could be something like:

```
[pantavisor] 11 INFO [Wed, 30 Aug 2023 20:49:34 +0000]    -- [config]: system.rundir = '/pv'
```
what do you think about it?

> Also, remember to add the new key to https://docs.pantahub.com/pantavisor-configuration/#pantavisorconfig once you are finished :)
Sure :)

asac:
imo the timestamp.format is an internal data thingy ... maybe its also a sink feature and you should have a per sink format as well ...  i am not keen to add more info to the log text in the internal data format thoughb

fortizc:
@asac @anibalportero so maybe something like `log.<output>.timestamp.format` where <output> could be filetree, singlefile or stdout. What do you think?

anibalportero:
> @asac @anibalportero so maybe something like `log.<output>.timestamp.format` where could be filetree, singlefile or stdout. What do you think?

I'm OK with that :)

PR #354, opened 2023-10-02
add hwrngd daemon with basic service/reaping infra in pv init

asac:


init.c
```diff
@@ -143,6 +143,117 @@ static int other_mounts()
...
+
+		if (!daemons[i]._respawning)
+			daemons[i]._respawning = 1;
+
+		daemons[i].pid = daemon_spawn(daemons);
+		pv_log(INFO, "spawned daemon %s: %d \n", daemons[i].name,
+		       daemons[i].pid);
+	next:
+		i++;
+	}
```
fortizc, line 211: I always tried to avoid "goto" so:
```suggestion
	for (int i = 0; daemons[i].name; ++i) {
		if (daemons[i].pid > 0 ||
		    (daemons[i].pid == 0 && !daemons[i].respawn))
			continue;

		if (stat(daemons[i].testpath, &sb)) {
			pv_log(INFO,
			       "daemon not enabled %s: disabling respawn\n",
			       daemons[i].name);
			daemons[i].pid = 0;
			daemons[i].respawn = 0;
			continue;
		}

		if (!daemons[i]._respawning)
			daemons[i]._respawning = 1;

		daemons[i].pid = daemon_spawn(daemons);
		pv_log(INFO, "spawned daemon %s: %d \n", daemons[i].name,
		       daemons[i].pid);
	}
```
Maybe there is no huge difference, but it feels better (at least to me)

anibalportero:
I'd say its a good candidate to be moved to its own daemons.h module :)

init.c
asac: ok but why ++i instead of i++ ?

fortizc: it's just habit

asac: did that in latest rev.

asac:
> I'd say its a good candidate to be moved to its own daemons.h module :)

I kinda agree, but then the init.c is about reaping processes and dealing with signal_handler which currently is too wired up with the daemons facility. maybe if we move the dropbear into this infra structure it would make sense (and getty?)....

anibalportero:
> > I'd say its a good candidate to be moved to its own daemons.h module :)
> 
> I kinda agree, but then the init.c is about reaping processes and dealing with signal_handler which currently is too wired up with the daemons facility. maybe if we move the dropbear into this infra structure it would make sense (and getty?)....

Both dropbear and getty are now in debug.c

asac:
> > I'd say its a good candidate to be moved to its own daemons.h module :)
> 
> I kinda agree, but then the init.c is about reaping processes and dealing with signal_handler which currently is too wired up with the daemons facility. maybe if we move the dropbear into this infra structure it would make sense (and getty?)....

ok i moved things to deamons.h etc. now....

PR #356, opened 2023-10-05
Feature/new disks stack

fortizc:
Adds a new disk implementation to support different disk types in a scalable way:
* Creates `disk_impl` to define a standard way to implement a disk type
* Rewrite the `pv_disk_mount_handler` to load the right implementation for each disk
* Modify the parser to support more fields
* Adds 2 different disk types: `volume-disk` and `swap-disk` (both applicable to zram)
* Rewrite the current crypt disk support in the new disk standard

Example of disk section in devices.json:
```
{
    "disks": [
        {
           "name": "dm-internal-secrets",
            "path": "/storage/dm-crypt-files/dm-internal-secrets/versatile.img,2,versatile_key-internal_secrets",
            "type": "dm-crypt-versatile"
        },
        {
            "name": "zram-swap-1",
	    "type": "swap-disk",
	    "format": "swap",
	    "provision": "zram",
            "provision_options": "disksize=1M"
        },
	{
            "name": "zram-swap-2",
	    "type": "swap-disk",
	    "format": "swap",
	    "provision": "zram",
            "provision_options": "disksize=1M comp_algorithm=lzo"
        },
	{
            "name": "zram-vol-1",
	    "type": "volume-disk",
	    "format": "ext4",
	    "provision": "zram",
            "provision_options": "disksize=1M",
	    "mount_target": "/zram_test"
        }
]

### CODE REVIEW

**Comment by asac** [comment]
Let's document the new disks.json spec with example in the PR description and lets make a matching patch for the docs/ repo with the latest format

File: atom.mk [hunk 1]
```diff
@@ -165,3 +171,15 @@ LOCAL_LIBRARIES := cryptsetup
...
+
+LOCAL_DESTDIR := ./
+LOCAL_MODULE := pvzram
+
+LOCAL_C_INCLUDES := $(LOCAL_PATH)
+
+LOCAL_SRC_FILES := disk/disk_zram_utils.c \
+		   utils/pv_zram.c
+
+include $(BUILD_EXECUTABLE)
```
anibalportero, line 185: what is this new executable for?

disk/disk.h
```diff
@@ -0,0 +1,116 @@
...
+	DISK_FORMAT_SWAP,
+	DISK_FORMAT_EXT4,
+	DISK_FORMAT_EXT3
+} pv_disk_format_t;
+
+typedef enum {
+	DISK_STATUS_ERROR,
+	DISK_STATUS_MOUNTED,
+	DISK_STATUS_NOT_MOUNTED
+} pv_disk_status_t;
```
anibalportero, line 50: we can move this to .c if we are not going to use it externally

state.c
```diff
@@ -444,6 +444,8 @@ int pv_state_validate(struct pv_state *s)
 
 static int pv_state_mount_bsp_volumes(struct pv_state *s)
 {
+	pv_disk_mount_all(s);
```
anibalportero, line 447: I think this kind of thing would break non-reboot updates.

pv_state_mount_bsp_volumes is about mounting the BSP volumes. That means mounting every volume that is not linked to a platform, such as the metadata volumes. It is called from pv_state_start, which is called from _pv_run in pantavisor.c and it is only called when we are not transitioning (non-reboot updates).

For disks, we had relied on the disk pointer that we store in each volume. When a platform is being started, we go through the list of volumes and call pv_volume_mount for the disks that are associated to each platform. Then, inside of pv_volume_mount we mount the disk if necessary with the handler, which was the only way to actually. And that's the only place we should be mounting disks, if they are required by a platform, that way, you can add a new platform that uses a new disk and everything work fine, because the work is done at the platform level basically. Same with umount.

disk/disk.h
```diff
@@ -0,0 +1,116 @@
...
+	char *provision_ops;
+	char *provision;
+	char *mount_target;
+	bool def;
+	bool mounted;
+	// pv_disk
+	struct dl_list list;
+};
+
+struct pv_disk *pv_disk_add(struct pv_state *s);
```
anibalportero, line 69: I might be wrong in this one but I think we should switch all functions in disk.h interface from working with pv_state into work with pv_disk in the same way we do with pv_volume in volumes.c, and leave all state level stuff to state.c . This works better in the context of the comment that I left in state.c and the spirit of working with disks individually instead of state-wide.

volumes.c
```diff
@@ -163,7 +161,7 @@ int pv_volume_mount(struct pv_volume *v)
 	}
 
 	if (v->disk && !v->disk->def && !v->disk->mounted) {
-		ret = pv_disks_mount_handler(v->disk, "mount");
+		ret = pv_disk_mount_handler(v->disk, "mount");
```
anibalportero, line 164: This is the same point where we should mount disks. Maybe changing the name of the function if handler is no longer used, then decide whether to use handler or not inside.

parser/parser_system1.c
```diff
@@ -256,36 +289,41 @@ static int parse_disks(struct pv_state *s, char *value)
...
-			d->type = DISK_UNKNOWN;
+		d->type = get_disk_type(tmp);
+		free(tmp);
+		if (d->type == DISK_UNKNOWN) {
+			pv_log(ERROR, "cannot add new disk, type = UNKNOWN");
 			goto out;
 		}
 
+		tmp = pv_json_get_value(str, "format", diskv, diskc);
+		d->format = get_disk_format(tmp);
```
anibalportero, line 325: Does DISK_FORMAT_UNKNOWN should also return error like DISK_UNKNOWN?

anibalportero (commented):
I like the new layout in terms of architecture. It makes it harder to justify a change as large as this one though, as things should be easier to add and modify in a modular way. This could have been added in different PRs to make implementation, reviewing and testing easier:
* the refactor of disks with crypt disk impl only
* swap disk
* volume disk

atom.mk
fortizc: This is a command line app to provisioning new zram disk. The idea is, for example, be able to add swap to a running device. This application use the same library used by the internal code.

I think I need to change this to be included in /usr/bin or any other directory in the PATH

disk/disk.h
fortizc: I will check it and move it if it's ok. Thanks!

disk/disk.h
fortizc: Yeah, that will be nice. Let me check if it's possible. I thought on this too, but at one point I forgot

disk/disk.h
fortizc: This is used in several places like disk_impl.h and each disk implementation

state.c
anibalportero: ack. If we are going to only mount swap disks in this point, I'd suggest to do it in a diff function named pv_state_mount_bsp_disks that is called before this one in pv_state_start

parser/parser_system1.c
```diff
@@ -256,36 +289,41 @@ static int parse_disks(struct pv_state *s, char *value)
 
 		if (jsmnutil_parse_json(str, &diskv, &diskc) <= 0) {
 			pv_log(ERROR, "Invalid disk entry");
-			continue;
+			goto next;
```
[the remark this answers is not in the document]
anibalportero: Yes, it's always better to alert the user something is bad rather than let it pass and work in an unexpected manner

disk/disk.c
```diff
@@ -0,0 +1,272 @@
...
+		} else {
+			impl = &swap_impl;
+		}
+		break;
+	case DISK_VOLUME:
+		if (!disk->provision) {
+			pv_log(ERROR,
+			       "cannot use disk, must define a provision");
+			return -1;
+#define PV_DISK_
```
anibalportero, line 172: is this a left over?

disk/disk.c
```diff
@@ -0,0 +1,272 @@
...
+
+	if (impl->umount(disk) != 0) {
+		pv_log(WARN, "cannot umount %s", disk->path);
+		return -1;
+	}
+
+	return 0;
+}
+
+int pv_disk_mount_handler(struct pv_disk *disk, const char *action)
```
anibalportero, line 145: I guess we could change the name of this function to pv_disk_mount or something more fit to what it does now

state.c
```diff
@@ -444,6 +444,8 @@ int pv_state_validate(struct pv_state *s)
 
 static int pv_state_mount_bsp_volumes(struct pv_state *s)
 {
+	pv_disk_mount_swap(s);
```
anibalportero, line 447: I'd move this out of here as it has nothing to do with mounting bsp volumes. You could do it before the call to pv_state_mount_bsp_volumes in pv_state_start.

@asac  would it be convenient to fail and return -1 if disks are not mounted, so we rollback in that case?

parser/parser_system1.c
```diff
@@ -230,6 +231,35 @@ static int parse_bsp_drivers(struct pv_state *s, char *v, int len)
 	return 1;
 }
 
+static pv_disk_format_t get_format(const char *str, jsmntok_t *diskv, int diskc)
```
anibalportero, line 234: maybe rename these to parse_disks_get_format, parse_disks_get_type... to get less generic names

parser/parser_system1.c
```diff
@@ -250,57 +280,54 @@ static int parse_disks(struct pv_state *s, char *value)
...
-		else {
-			d->type = DISK_UNKNOWN;
+		if (d->type == DISK_UNKNOWN) {
+			pv_log(ERROR, "cannot add new disk, type = UNKNOWN");
 			goto out;
 		}
 
-		free(tmp);
-		tmp = NULL;
+		d->format = get_format(str, diskv, diskc);
```
anibalportero, line 318: I'd go to out in case this returns DISK_UNKNOWN

disk/disk.h
fortizc: Check https://github.com/pantavisor/pantavisor/pull/363

disk/disk.c
fortizc: Removed in https://github.com/pantavisor/pantavisor/pull/363

disk/disk.c
fortizc: This was reimplemented to improve the name :) and avoid passing the action as string. See https://github.com/pantavisor/pantavisor/pull/363

parser/parser_system1.c
fortizc: Added in https://github.com/pantavisor/pantavisor/pull/363

PR #361, opened 2023-10-24
improve waitpids to only reap valid pids for daemons

asac:


fortizc:
This PR do not fix the problem at least in x64-ubuntu

asac:
@fortizc please describe the problem you are referring to?

PR #363, opened 2023-10-31
Bugfix/disk review

fortizc:
This PR address several problems/improvements mentioned in the original disk PR:
* Replace API to use disk (`struct dl_list *disks`) instead of `pv_state *s`
* Remove some confusing logs.
* Remove unused `#define`
* Change API to mount/umount disk
* Rename disk parse functions
* Adds checks to detect unknown formats in the parser

### DIFF
--- a/state.c
+++ b/state.c
@@ -123,7 +123,7 @@ void pv_state_free(struct pv_state *s)
 	pv_drivers_empty(s);
 	pv_platforms_empty(s);
 	pv_volumes_empty(s);
-	pv_disk_empty(s);
+	pv_disk_empty(&s->disks);
 	pv_addons_empty(s);
 	pv_objects_empty(s);
 	pv_jsons_empty(s);
@@ -444,7 +444,8 @@ int pv_state_validate(struct pv_state *s)
 
 static int pv_state_mount_bsp_volumes(struct pv_state *s)
 {
-	pv_disk_mount_swap(s);
+	if (pv_disk_mount_swap(&s->disks) != 0)
+		return -1;
 
 	struct pv_volume *v, *tmp;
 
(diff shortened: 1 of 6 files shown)


state.c
```diff
@@ -444,7 +444,7 @@ int pv_state_validate(struct pv_state *s)
 
 static int pv_state_mount_bsp_volumes(struct pv_state *s)
 {
-	pv_disk_mount_swap(s);
+	pv_disk_mount_swap(&s->disks);
```
anibalportero, line 447: still think that we should return -1 here in case swap is specified but the mount failed for some reason

anibalportero (commented):
resolved all comments from [356](https://github.com/pantavisor/pantavisor/pull/356) except one

state.c
fortizc: OK, but currently the function `pv_disk_mount_swap()` only fails if the disks list is empty and only show a log message when a swap disk cannot be mounted.
```C
int pv_disk_mount_swap(struct dl_list *disks)
{
	if (!disks)
		return -1;

	pv_log(INFO, "mounting all swap disk");

	struct pv_disk *d, *tmp;
	dl_list_for_each_safe(d, tmp, disks, struct pv_disk, list)
	{
		if (d->type != DISK_SWAP)
			continue;

		int err = pv_disk_mount(d);
		if (err != 0)
			pv_log(ERROR, "cannot mount %s", d->name);
	}

	return 0;
}
```
So, if I understand right and trying to give sense to this, the function should return -1 if any disk fails?

anibalportero: if something is in the specification and it fails for whatever reason, we are not running what the specification contains. So yes, we should return -1 if something fails in pv_disk_mount_swap and then return -1 in pv_state_mount_bsp_volumes too. That way we'll get a rollback in case a disk mount fails.

fortizc: fixed :)

PR #371, opened 2023-11-14
Rework init reaping code so it reaps only direct childs.

asac:
* introduce a background pid pool so we can reap them without stopping pantavisor
 * introduce tsh facilities to push and pop tracked pids
 * remember forked lxc processes that exit after daemonizing as background ids
 * remember tsh_run pids started in background in pid pool
 * also fix potential issues in run processes by resetting sigmask in child

### DIFF
--- a/init.c
+++ b/init.c
@@ -144,17 +144,6 @@ static int other_mounts()
 	return 0;
 }
 
-static pid_t waitpids(int *wstatus)
-{
-	pid_t pid;
-
-	if ((pid = waitpid(-1, wstatus, WNOHANG)) > 0) {
-		return pid;
-	}
-
-	return 0;
-}
-
 static void signal_handler(int signal)
 {
 	pid_t pid = 0;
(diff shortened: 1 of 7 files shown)


init.c
```diff
@@ -147,26 +147,11 @@ static int other_mounts()
 static pid_t waitpids(int *wstatus)
```
fortizc, line 147: This function is so small, maybe we can just remove it and call `waitpid(-1....` inside of the handler

utils/tsh.c
```diff
@@ -65,6 +65,40 @@ static char **_tsh_split_cmd(char *cmd)
 	return ts;
 }
 
+pid_t bgpids[128];
```
fortizc, line 68: The size (128) is used in both of the functions below, it should be a macro

init.c
```diff
@@ -190,8 +175,10 @@ static void signal_handler(int signal)
 			       "Respawn of critical service failed %d: %s", pid,
 			       strerror(errno));
 		}
-		if (pv_pid == 0 || (pv_debug_is_ssh_pid(pid)))
+		if (getpid() == 1 || pv_debug_is_ssh_pid(pid) || tsh_bgid_pop(pid))
```
fortizc, line 178: The comment of this line is at the beginning of the loop (line 167) but it's outdated and incomplete

init.c
anibalportero: done

utils/tsh.c
anibalportero: done

init.c
anibalportero: done

PR #373, opened 2023-11-15
fix logserver pid reaping causing double pv_stop

asac:


anibalportero:
@asac is this one still relevant?

PR #388, opened 2023-11-24
Adds config key to setup a max disk usage to log without compress

fortizc:
Adds `log.maxsize.storage` this key establishes a limit of disk usage, if that limit is surpassed then the logserver will compress the logs.


anibalportero:
Instead of creating a new key, we could reuse log.maxsize and make it behave like storage.logtempsize, where you can set a % suffix at the end so it takes it as a percentage, if that's the idea

fortizc:
> Instead of creating a new key, we could reuse log.maxsize and make it behave like storage.logtempsize, where you can set a % suffix at the end so it takes it as a percentage, if that's the idea

@anibalportero but log.maxsize refers to the log size and this new key refers to the percentage of free space on disk

anibalportero:
> > Instead of creating a new key, we could reuse log.maxsize and make it behave like storage.logtempsize, where you can set a % suffix at the end so it takes it as a percentage, if that's the idea
> 
> @anibalportero but log.maxsize refers to the log size and this new key refers to the percentage of free space on disk

Then I guess we could name it something like log.maxsize.threshold so it resemble storage.gc.threshold? That's if they do something similar, which I think they do?

fortizc:
> > > Instead of creating a new key, we could reuse log.maxsize and make it behave like storage.logtempsize, where you can set a % suffix at the end so it takes it as a percentage, if that's the idea
> > 
> > 
> > @anibalportero but log.maxsize refers to the log size and this new key refers to the percentage of free space on disk
> 
> Then I guess we could name it something like log.maxsize.threshold so it resemble storage.gc.threshold? That's if they do something similar, which I think they do?

Done!

PR #402, opened 2024-01-15
rpiab booloader support for Raspbery Pi Firmware A/B tryboot

asac:
rpiab booloader support for Raspbery Pi Firmware A/B tryboot
    
     * ship pantavisor and kernel as part of full raspberry pi boot partition
     * install dtbs to boot partition from kernel
     * use mcopy to copy autoboot.txt to and from partition 1
     * patch cmdline.txt in boot-a/b partition to have pv_rev
     * add pvzlib.c/h helper to support shipping gzipped boot partitions
     * add support for shipping 'rpiab' partition in run.json
     * parse /proc-device-tree and autoconf.txt in early boot to find tryboot
       info and partition and boot disk (through boot-mode)
     * set tryboot flag through IOCTL_MBOX_PROPERTY ioctl (mbox_property)
    
    Example run.json:
    {
        "addons": [],
        "firmware": "firmware.squashfs",
        "initrd_config": "trail.config",
        "modules": "modules.squashfs",
        "rpiab": "pantavisor-rpi.img.gz"
    }
    
    See: https://www.raspberrypi.com/documentation/computers/raspberry-pi.html#fail-safe-os-updates-tryboot
    See: https://www.raspberrypi.com/documentation/computers/config_txt.html#autoboot-txt
    See: https://github.com/raspberrypi/userland/tree/master/host_applications/linux/apps/vcmailbox

### DIFF
--- a/atom.mk
+++ b/atom.mk
@@ -16,7 +16,7 @@ include $(BUILD_SHARED_LIBRARY)
 
 include $(CLEAR_VARS)
 
-LOCAL_LIBRARIES := libthttp mbedtls picohttpparser
+LOCAL_LIBRARIES := libthttp mbedtls picohttpparser zlib
 LOCAL_CONDITIONAL_LIBRARIES := OPTIONAL:e2fsprogs
 
 LOCAL_DESTDIR := ./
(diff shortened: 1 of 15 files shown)


bootloader.h
```diff
@@ -46,6 +53,10 @@ int pv_bootloader_set_installed(char *rev);
 int pv_bootloader_set_commited(char *rev);
 int pv_bootloader_set_failed(void);
 
+int pv_bootloader_install_update(struct pv_update *update);
```
anibalportero, line 56: why not just using pv_bootloader_set_installed and pv_bootloader_set_commited? aren't we using these along the old ones?

anibalportero (commented):
the most important thing here is that the new bootloader functions are making all updates to fail. Maybe we should just reuse (maybe rename too) the old functions and use the new ones at the beginning if the new funcitons pointers are set

PR #403, opened 2024-01-22
fix: dmverity and crypt disks not being preemptively closed

anibalportero:
It has been observed in App Engine that, if Pantavisor cannot properly shutdown, it would leave remnants for the next execution. This would make Pantavisor fail unless the host machine were rebooted.

With this PR, we want to preemptively close those crypt and dmverity vols before trying to initialize them.

### DIFF
--- a/scripts/volmount/crypt/crypt
+++ b/scripts/volmount/crypt/crypt
@@ -160,6 +160,24 @@ do_dm_sanity()
 	which mkfs.ext4 &> /dev/null || exit 111
 }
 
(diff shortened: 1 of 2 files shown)


asac:
looks good in general. i would make the close dependent on whether the device exists and not have the || true only. Also log that this decision was taken?

anibalportero:
> looks good in general. i would make the close dependent on whether the device exists and not have the || true only. Also log that this decision was taken?

done

PR #411, opened 2024-03-20
fix: rpiab bootloader support breaking other bootloaders

anibalportero:
This PR contains the corrections for https://github.com/pantavisor/pantavisor/pull/402:

* bootloader: use a single set of install, commit and fail functions in bootloader.h
* bootloader: fix install failing in case the new bootloader semantic is not being used
* parser: make parsing fail in case bootimg is not specified and we are expecting it by config
* rpiab: remove pv_update and pv_state from interfaces. Only used to check bootimg not set, which was moved to parser (see previous bullet). Changed in favor of char *rev to make it more similar to the rest of the bootloader types
* state: free bootimg in pv_state_free
* updater: call bootloader installed and commit functions only once. In the case of pv_bootloader_commit_update, it has been changed to the old place. IMO it makes no sense to report DONE to cloud and save done in /storage if the bootloader failed to commit and thus the whole update will fail for Pantavisor

### DIFF
--- a/bootloader.c
+++ b/bootloader.c
@@ -130,30 +130,42 @@ bool pv_bootloader_trying_update()
 			 strlen(pv_bootloader_get_rev()) + 1));
 }
 
-int pv_bootloader_set_installed(char *rev)
+int pv_bootloader_install_update(char *rev)
 {
 	pv_log(INFO,
 	       "setting installed revision %s to be started after next reboot",
 	       rev);
+
+	if (ops->install_update)
+		return ops->install_update(rev);
+
 	return pv_bootloader_set_try(rev);
 }
 
-int pv_bootloader_set_commited(char *rev)
+int pv_bootloader_commit_update(char *rev)
 {
 	if (!ops)
 		return -1;
 
 	pv_log(INFO, "setting done revision %s to be started after next reboot",
 	       rev);
+
+	if (ops->commit_update)
+		return ops->commit_update();
+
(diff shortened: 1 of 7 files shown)


updater.c
```diff
@@ -1420,15 +1420,14 @@ int pv_update_finish(struct pantavisor *pv)
...
+		if (pv_bootloader_commit_update(pv->state->rev)) {
 			pv_log(ERROR,
-			       "revision for next boot could not be set");
+			       "revision could not be committed to bootloader");
 			ret = -1;
 			goto out;
 		}
 		pv_update_set_status(u, UPDATE_DONE);
 		pv_storage_set_rev_done(pv->state->rev);
-		pv_bootloader_commit_update();
```
asac, line 1431: @anibalportero lets not remove the two teps for committing. i had thought a while about this for rpiab case and did it so we are safe in case reboot happen between the set_committed and the set_rev

bootloader.c
```diff
@@ -130,30 +130,42 @@ bool pv_bootloader_trying_update()
...
 		return -1;
 
 	pv_log(INFO, "setting done revision %s to be started after next reboot",
 	       rev);
+
+	if (ops->commit_update)
+		return ops->commit_update();
+
 	return (pv_bootloader_set_rev(rev) || pv_bootloader_unset_try() ||
 		ops->flush_env());
```
asac, line 157: in the previous code this was also run for rpiab update... before the commit_update. now we dont run this anymore?

PR #416, opened 2024-03-24
support modules_$uts modules in run.json

asac:


anibalportero:
It would be nice to have this documented in https://docs.pantahub.com/pantavisor-state-format-v2/#bsprunjson

PR #417, opened 2024-03-28
Fixes detection of ubifs volumes

fortizc:
Adds to blkid lib the ubifs detection using /sys filesystem:

1. Checks the fs with pv_config_get_storage_fstype()
2. Gets the device and volume names from pv_config_get_storage_path(), normally in the format device:volume (e.g. ubi0:pvroot)
3. Checks the device in /sys/devices/virtual/ubi
4. Checks all volumes names, if the volume it's found, then the path in format /dev/<device> will be written in the blkid_info struct

### DIFF
--- a/blkid.c
+++ b/blkid.c
@@ -35,11 +35,14 @@
 #include <errno.h>
 #include <string.h>
 #include <stdint.h>
+#include <linux/limits.h>
 
 #include "blkid.h"
 #include "utils/math.h"
 #include "str.h"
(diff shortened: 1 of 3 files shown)


asac:
looks good.

anibalportero:
We should document this in https://docs.pantahub.com/storage/ (maybe adding a new file system type section in there) and also in https://docs.pantahub.com/pantavisor-configuration/ (just the new syntax)

PR #423, opened 2024-05-07
Fix driver loading/unloading

fortizc:
Fixes the unload/load problem, setting the `loaded` field in the `pv_platform_driver` struct


anibalportero (commented):
I'm not sure what is the problem, but d->loaded seems to be OK? according to the issue description, the GET request is returning LOADED, which seems good and that's taken from d->loaded I think.

One thing that needs fixing is that we are not handling the output of pv_platform_load_drivers and pv_platform_unload_drivers in ctrl.c. We should do that to the possible extend. Then, reproduce the issue and see what is that returning.

PR #425, opened 2024-05-14
Fixes problems found to update the toolchain

fortizc:
* Remove unnecessary fclose(), on rpiab.c:239
* Remove an "always true" condition , on updater.c:466 

Both are warning (and errors due to -werror) with the new compiler

### DIFF
--- a/updater.c
+++ b/updater.c
@@ -456,7 +456,7 @@ static char *pv_update_get_progress_json(struct pv_update_progress *progress)
 		pv_json_ser_string(&js, progress->msg);
 		pv_json_ser_key(&js, "progress");
 		pv_json_ser_number(&js, progress->progress);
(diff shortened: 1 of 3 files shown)


updater.c
```diff
@@ -456,7 +456,7 @@ static char *pv_update_get_progress_json(struct pv_update_progress *progress)
 		pv_json_ser_string(&js, progress->msg);
 		pv_json_ser_key(&js, "progress");
 		pv_json_ser_number(&js, progress->progress);
-		if (progress->data && (strlen(progress->data) > 0)) {
+		if (strlen(progress->data) > 0) {
```
asac, line 459: whats wrong with checking for progress->data != NULL?

fortizc: because is defined here https://github.com/pantavisor/pantavisor/blob/master/updater.c#L263 as a fixed size array

PR #426, opened 2024-05-17
allow quotes to accept special characters in cmdline values

anibalportero:
This PR contains a modification of the cmdline parser to support quoted configuration arguments.

Syntax rules have been kept as similar as the Linux Kernel cmdline as possible:

* Any number of spaces between items is allowed.
* The ' ' special character is used as a separator between items.
* The '"' special character is used to escape separators, using two between each config item or between the value.
* The '"' between items, and values are discarded after parsing the cmdline.
* The '=' special character cannot be escaped.
* The first '=' found separates the key and the value of an item, while the rest of '=' are part of the value.
* Empty keys are not allowed.

While keeping Pantavisor special rules for backwards compatibility with the current configuration syntax:

* Items without '=' are ignored.
* The '\n' special character is used as an equivalent to the ' ' separator.
* Items need to have a prefix ("pv_" or "ph_") to be acknowledged by Pantavisor.
* Empty values are allowed and substituted by an empty string.
* The processed keys depend on the set of Pantavisor configuration keys.

### DIFF
--- a/config.c
+++ b/config.c
@@ -262,20 +262,19 @@ bool pv_config_get_bool(config_index_t ci)
 	return entries[ci].value.b;
 }
 
-static void _set_config_by_entry_bool(struct pv_config_entry *entry, bool value,
-				      level_t modified)
+static void _set_config_by_entry_bool(struct pv_config_entry *entry, bool value)
 {
 	if (!entry)
 		return;
 
 	entry->value.b = value;
-	entry->modified = modified;
 }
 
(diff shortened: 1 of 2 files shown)


fortizc:
LGTM 👍🏼

asac:
where does our parser differ from what the kernel parses from cmdline to ENV? the closer we are to that behaviour the better we are prepared to later move to envs from this approach.

PR #428, opened 2024-05-23
Draft: introduce skeleton code for system.init.mode == IM_INSTALLER

asac:


mount.c
```diff
@@ -90,6 +90,22 @@ void pv_mount_umount(void)
 	}
 }
 
+static int pv_mount_install(struct pv_init *this)
+{
+	char path[PATH_MAX];
+	struct stat st;
+	sprintf(path, "/lib/pv/pantavisor-installer");
```
anibalportero, line 97: setting this in paths.h would be the way to go as "/lib" mount point is configurable

asac: yeah i just moved this patch here... think this will be re-designed as part of firstboot install mode...

PR #429, opened 2024-05-27
env level configuration support

anibalportero:
This change adds Linux environment variables as an additional level of configuration between cmdline (deprecated from now on) and trail config. It also adds a fully new set of config key to make its syntax compatible with env, while keeping backwards compatibility of the old config keys.

These are the main points of action:

* New env friendly config key format, making it case insensitive and eliminating "." characters. For example, "debug.ssh" would be now "debug_ssh" or "DEBUG_SSH".
* Add new aliases to keep backwards compatibility. For example, "debug.ssh" would be still supported but deprecated.
* Keep the old format for pvcontrol config ls. A new command will be issued in the future with the new one while keeping support for the old one.
* Change all index names to be equivalent to the upper case version of the new keys.

TODO:
* Set configuration from Linux environment variables, following the new key syntax format.
* What to do about sysctl config keys? We cannot switch "_" with "/" in this case.
* Test and debug.
* Documentation.


asac:
Looks promising.

asac:
For sysctl we might need to stick to old way for now. And later introduce a sysctl.XXXX.XXXX.XXX parsing option (without pv_ prefix) which we would offer for kernels < 5.8 (since 5.8 kernel will parse such cmdline options directly). We wont need them in envs then anymore. By offering sysctl... option will help folks transition to the pure kernel approach without pantavisor ....

PR #432, opened 2024-05-28
allow sysctl config from env

anibalportero:
This PR adds the sysctl config keys to the [Linux env configurable list](https://github.com/pantavisor/pantavisor/pull/429):

* The old format (e.g: "sysctl.kernel.core_pattern") is still supported.
* A new env friendly format (e.g: "PV_SYSCTL_KERNEL_CORE_PATTERN") is now supported. This is parsed by checking for existing dirs when reaching a '_' character. If true, then it is switched with a '/'.

### DIFF
--- a/config.c
+++ b/config.c
@@ -79,6 +79,8 @@ typedef enum {
 #define UPDATE_TIME POLICY | TRAIL
 #define RUN_TIME UPDATE_TIME | META | CMD
 
+#define LEVEL_SYSCTL PV_ENTRY | UPDATE_TIME
+
 // default list
 #define CACHE_DEVMETADIR_DEF "/storage/cache/devmeta"
 #define CACHE_USRMETADIR_DEF "/storage/cache/meta"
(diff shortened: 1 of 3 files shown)


config_parser.c
```diff
@@ -283,3 +284,69 @@ void config_clear_items(struct dl_list *list)
...
+{
+	const char *start = key + strlen("sysctl");
+	char *path =
+		calloc(strlen("/proc/sys") + strlen(start) + 1, sizeof(char));
+	if (!path)
+		return NULL;
+
+	sprintf(path, "%s", "/proc/sys");
+
+	const char *next = start;
```
asac, line 298: isnt next/start redundant?

config.c
```diff
@@ -696,30 +696,16 @@ static char *_get_value_policy(struct dl_list *config_list)
...
-	for (int i = 0; i < (int)strlen(start); ++i)
-		p[i] = next[i] == '.' ? '/' : next[i];
-
-	errno = 0;
 	int fd = open(path, O_WRONLY | O_SYNC);
 	if (fd < 0) {
-		pv_log(ERROR, "open failed for sysctl node %s with '%s'", path,
-		       strerror(errno));
-
+		pv_log(ERROR, "cannot open '%s': %s", path, strerror(errno));
```
asac, line 708: since we seem to ignore the errors in the iterator, maybe we should make this a WARN?

asac:
lgtm if it works for the core_pattern case :)

PR #436, opened 2024-06-21
OEM signature truststore

anibalportero:
This PR is divided in two commits:
* a small refactor of the pv_signature_parse_certs to divide it in several more manageable functions
* the new feature to enable a new certifiacte truststore location in Pantavisor for OEM providers

The new feature can be used playing with these two configuration keys
* PV_SECUREBOOT_OEM_CN_NAME: with this string, we can set the certificate subject CN name that Pantavisor will expect in the secureboot signature file to use the alternative OEM certificate truststore for validation. If that CN is not found in the signature file, it fallback to the default certificate truststore
* PV_SECUREBOOT_OEM_TRUSTORE: the equivalent to PV_SECUREBOOT_TRUSTORE for OEM


signature.c
```diff
@@ -603,7 +605,25 @@ static int pv_signature_parse_certs(struct dl_list *certs_raw, struct mbedtls_x5
...
+				const char *path)
+{
+	config_index_t name = PV_SECUREBOOT_TRUSTSTORE;
+
+	char oem_name[256];
+	SNPRINTF_WTRUNC(oem_name, 256, "oem-%s",
+			pv_config_get_str(PV_SECUREBOOT_OEM_ID));
+
+	// subject CN is taken from the first certificate of the chain
+	if (pv_str_matches(oem_name, strlen(oem_name), certs->subject.val.p,
```
asac, line 618: is .p the parsed CN= field of subject? or the whole Subject? if not we need to get the CN= field from that subject and compare that one i believe

paths.h
```diff
@@ -150,6 +150,7 @@ void pv_paths_writable(char *buf, size_t size);
 #define PV_DEFAULTS_GROUPS PV_DNAME "/defaults/groups.json"
 #define PVS_TRUST_DNAME PV_DNAME "/pvs/trust"
 #define PVS_PK_FNAME PVS_TRUST_DNAME "/pub.pem"
+#define PVS_CERT_OEM_STORE "oem-ca-certificates"
```
asac, line 153: is this PVS_CERT_DEFAULT_OEM_STORE?

anibalportero: that's right

signature.c
anibalportero: seems like subject is a list, so we have to check the "CN" in subject.oid and then get the value from subject.val

paths.h
```diff
@@ -150,6 +150,7 @@ void pv_paths_writable(char *buf, size_t size);
 #define PV_DEFAULTS_GROUPS PV_DNAME "/defaults/groups.json"
 #define PVS_TRUST_DNAME PV_DNAME "/pvs/trust"
 #define PVS_PK_FNAME PVS_TRUST_DNAME "/pub.pem"
+#define PVS_CERT_DEFAULT_OEM_STORE "oem-ca-certificates"
```
asac, line 153: this should be now ca-oem-certificates

PR #437, opened 2024-06-28
OEM configuration level

anibalportero:
This PR is a continuation of the [OEM signature trustore](https://github.com/pantavisor/pantavisor/pull/436).

In order to keep the configuration as small as possible, `PV_SECUREBOOT_OEM_CN_NAME` is now named `PV_OEM_NAME`. The same functionality for secureboot from the previous PR is kept.

This new `PV_OEM_NAME` can also be used to load additional configuration from containers, with the following behavior, when defined:
* If PV_POLICY is set, the file is loaded from the trail file <PV_OEM_NAME>/<PV_POLICY>.config
* If PV is not set, the file is is loaded from <PV_OEM_NAME>/default.config
* If the configured file is not found, we continue normally with a log WARN.

Additionally, trail config (now, bsp config) is now done after signature validation and before oem config.

All config levels have been reviewed with the following criteria:
* All PH_* keys are now allowed in the update level (oem config or bsp config). For any misconfiguration at this level, we rely on the fallback mechanism when ph connection is not possible.
* The PV_CONTROL_* keys are not allowed in the update level. In this case, we still need to implement a better system to prevent device control loss.
* The PV_WDT_* keys are not allowed in the update level.

### DIFF
--- a/config.c
+++ b/config.c
@@ -69,17 +69,18 @@ typedef enum {
 	PV_CMDLINE = 1 << 5,
 	PH_CMDLINE = 1 << 6,
 	ENV = 1 << 7,
-	OEM = 1 << 8,
-	META = 1 << 9,
-	CMD = 1 << 10
+	BSP = 1 << 8,
+	OEM = 1 << 9,
+	META = 1 << 10,
+	CMD = 1 << 11
 } level_t;
 
-#define PV_ENTRY PV_CONF | PV_CMDLINE | ENV
-#define PH_ENTRY PH_CONF | PH_CMDLINE | ENV
-#define UPDATE_TIME POLICY | OEM
-#define RUN_TIME UPDATE_TIME | META | CMD
+#define PV PV_CONF | POLICY | PV_CMDLINE | ENV
+#define PH PH_CONF | POLICY | PH_CMDLINE | ENV
+#define UPDATE BSP | OEM
+#define RUN META | CMD
 
-#define LEVEL_SYSCTL PV_ENTRY | UPDATE_TIME
+#define LEVEL_SYSCTL PV | UPDATE
 
 // default list
 #define CACHE_DEVMETADIR_DEF "/storage/cache/devmeta"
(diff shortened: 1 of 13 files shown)


config.c
```diff
@@ -120,140 +121,106 @@ struct pv_config_entry {
...
 	  .value.s = SECUREBOOT_OEM_TRUSTSTORE_DEF },
-	{ STR, "PV_SECUREBOOT_TRUSTSTORE", PV_ENTRY | POLICY, 0,
+	{ STR, "PV_SECUREBOOT_TRUSTSTORE", PV, 0,
 	  .value.s = SECUREBOOT_TRUSTSTORE_DEF },
-	{ STR, "PV_STORAGE_DEVICE", PV_ENTRY | POLICY, 0, .value.s = NULL },
-	{ STR, "PV_STORAGE_FSTYPE", PV_ENTRY | POLICY, 0, .value.s = NULL },
-	{ BOOL, "PV_STORAGE_GC_KEEP_FACTORY", PV_ENTRY | RUN_TIME, 0,
+	{ STR, "PV_STORAGE_DEVICE", PV, 0, .value.s = NULL },
+	{ STR, "PV_STORAGE_FSTYPE", PV, 0, .value.s = NULL },
+	{ BOOL, "PV_STORAGE_GC_KEEP_FACTORY", PV | UPDATE | RUN, 0,
```
asac, line 197: Can we name UPDATE == OEM? or is that used elswhere where UPDATE makes more sense?

anibalportero: UPDATE is OEM + BSP (trail.config), it's only a define to save characters in the table

anibalportero:
Will land this. With a couple of things pending for further PRs:
* Decide whether we remove trail.config or not
* Enable changing PV_CONTROL_ in OEM config. For that, we need to avoid remote revisions going local from OEM config

config.c
```diff
@@ -954,18 +923,74 @@ static int pv_config_load_creds_from_file(char *path)
...
+	config_iterate_items(&oem_config_list, _set_config_by_key,
 			     (void *)&level);
 
-	config_clear_items(&trail_config_list);
+	config_clear_items(&oem_config_list);
+
+	return 0;
+}
+
+static int pv_config_load_bsp(const char *rev, const char *trail_config)
```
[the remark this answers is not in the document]
anibalportero: Moving this to a further landing. For now, we keep it supported

PR #441, opened 2024-07-25
WIP: Feature/new hub logger

fortizc:
Adds new phlogger service.
* Use a queue (on a file) to temporary store logs
* Receive logs in json format
* Stop using xattr to check the already sent logs


utils/pvnanoid.c
```diff
@@ -0,0 +1,96 @@
...
+{
+	int fd = open("/dev/urandom", O_RDONLY);
+	if (fd < 0)
+		return -1;
+
+	ssize_t nbytes =
+		pv_fs_file_read_nointr(fd, nid->buf, PV_NANOID_RNG_BUF_SIZE);
+
+	if (nbytes < 0) {
+		nid->ok = false;
```
asac, line 46: feels like this one should close (fd) too

fortizc: you're right!

phlogger/phlogger.c
```diff
@@ -0,0 +1,580 @@
...
+	ssize_t nread = 0;
+	char *line = NULL;
+	char *logs = strdup("[");
+	size_t total_size = 1;
+	int i = 0;
+
+	rewind(f);
+
+	while ((nread = getdelim(&line, &len, PHLOGGER_LOG_DELIM, f)) != -1 &&
+	       i < PHLOGGER_MAX_LOG_SENT) {
```
anibalportero, line 384: it might be interesting to make this configurable if that is possible

anibalportero:
Please, add more info to the description with a high level explanation on how the new service works etc

PR #446, opened 2024-08-14
Sets automatically ignore_loglevel and printk_devkmsg=on when is needed

fortizc:
This PR allow to logserver set `ignore_loglevel` and `printk_devkmsg=on` when is needed, this is when some of the stdout* outputs is active (both options in this case) or when pv_capture_dmesg is true (only `printk_devkmsg=on`)

### DIFF
--- a/logserver/logserver.c
+++ b/logserver/logserver.c
@@ -868,7 +868,14 @@ void pv_logserver_toggle(struct pantavisor *pv, const char *running_rev)
 
 static void logserver_capture_dmesg()
 {
-	errno = 0;
+	if ((logserver.active_out & LOG_SERVER_OUTPUT_STDOUT) ||
+	    (logserver.active_out & LOG_SERVER_OUTPUT_STDOUT_PANTAVISOR))
+		return;
+
+	if (logserver_utils_printk_devmsg_on() == 0)
+		pv_log(DEBUG,
+		       "setting printk_devmsg=on, all message will be captured");
+
 	int fd = open("/dev/kmsg", O_RDONLY | O_NONBLOCK);
 	if (fd < 0) {
 		pv_log(WARN, "could not susbcribe dmesg to logserver %s",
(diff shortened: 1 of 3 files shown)


logserver/logserver.c
```diff
@@ -868,7 +868,13 @@ void pv_logserver_toggle(struct pantavisor *pv, const char *running_rev)
 
 static void logserver_capture_dmesg()
 {
-	errno = 0;
+	if (logserver_utils_printk_devmsg_on() < 0)
+		pv_log(DEBUG,
+		       "could not set printk_devmsg=on, some messages could lose");
```
anibalportero, line 873: you mean "some messages could be lost"?

logserver/logserver.c
```diff
@@ -915,6 +921,20 @@ int pv_logserver_init(const char *rev)
 		logserver_load_outputs();
 	}
 
+	if ((logserver.active_out & LOG_SERVER_OUTPUT_STDOUT) ||
+	    (logserver.active_out & LOG_SERVER_OUTPUT_STDOUT_DIRECT) ||
+	    (logserver.active_out & LOG_SERVER_OUTPUT_STDOUT_CONTAINERS) ||
+	    (logserver.active_out & LOG_SERVER_OUTPUT_STDOUT_PANTAVISOR)) {
+		if (logserver_utils_ignore_loglevel() != 0)
+			pv_log(DEBUG,
+			       "stdout: couldn't disable kernel log level, some messages could lose");
```
anibalportero, line 930: same

logserver/logserver.c
```diff
@@ -868,7 +868,13 @@ void pv_logserver_toggle(struct pantavisor *pv, const char *running_rev)
 
 static void logserver_capture_dmesg()
 {
-	errno = 0;
+	if (logserver_utils_printk_devmsg_on() < 0)
+		pv_log(DEBUG,
+		       "could not set printk_devmsg=on, some messages could lose");
+	else
```
anibalportero, line 874: I'd remove this and maybe other new log DEBUG traces you put there except the ones that indicate an error to make it less verbose. But if you think they are useful for device operation, we can keep them.

logserver/logserver.c
fortizc: 🤦🏼

PR #457, opened 2024-11-18
fix: ph logger not working with tmpfs logs

anibalportero:
This PR fixes ph_logger not being compatible with tmpfs (because of xattr).

The code in charge of managing the offset of the devices, which relied completely on writing/reading xattr from files,  has been moved to two functions:

*  _save_log_file_pos: it saves the offset on memory and as an xattr in files if possible by config.
* _load_log_file_pos: it loads the offset from memory, of not available from xattr, it takes 0 as the initial point.

The offsets are stored in memory as a list of paths and position offsets in ph_logger struct.

### DIFF
--- a/ph_logger.c
+++ b/ph_logger.c
@@ -88,6 +88,12 @@ struct ph_logger_fragment {
 
 static DEFINE_DL_LIST(frag_list);
 
+struct ph_logger_file {
+	char *path;
+	off_t pos;
+	struct dl_list list; // ph_logger_file
+};
+
 struct ph_logger {
 	int flags;
 	int epoll_fd;
@@ -97,6 +103,7 @@ struct ph_logger {
 	pid_t log_service;
 	pid_t range_service;
 	pid_t push_service;
+	struct dl_list files; // ph_logger_file
 };
 
 static struct ph_logger ph_logger = { .epoll_fd = -1,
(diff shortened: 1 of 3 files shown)


ph_logger.c
```diff
@@ -222,6 +229,74 @@ static char *strnchr(char *src, char ch, int len)
...
+static void _save_log_file_pos(off_t pos, const char *path)
+{
+	// try to get file info from memory
+	struct ph_logger_file *f;
+	f = _search_log_file(path);
+	// if not, we create a new entry
+	if (!f) {
+		pv_log(DEBUG, "log file '%s' not yet stored in memory. Saving new file...", path);
+		f = calloc(1, sizeof(struct ph_logger_file));
+		f->path = strdup(path);
```
fortizc, line 253: A memory check is needed here

ph_logger.c
```diff
@@ -222,6 +229,74 @@ static char *strnchr(char *src, char ch, int len)
...
+	// first, try to get pos from memory
+	struct ph_logger_file *f;
+	f = _search_log_file(path);
+	if (f)
+		return f->pos;
+
+	pv_log(DEBUG, "unknown file found in '%s'", path);
+
+	// if not in memory, try to get it from xattr if possible
+	char dst[32] = { 0 };
```
fortizc, line 282: This "32" is, I guess, the maximum size of the xattr. Maybe is good Idea to create a `#define [name] 32` to avoid errors, because this value is used below

ph_logger.c
```diff
@@ -222,6 +229,74 @@ static char *strnchr(char *src, char ch, int len)
...
+		pv_log(DEBUG, "log file '%s' not yet stored in memory. Saving new file...", path);
+		f = calloc(1, sizeof(struct ph_logger_file));
+		f->path = strdup(path);
+		dl_list_add(&ph_logger.files, &f->list);
+	}
+	if (!f)
+		pv_log(ERROR, "could not initialize log file: %s", strerror(errno));
+
+	// update pos value in memory
+	f->pos = pos;
```
fortizc, line 260: This assignment could be an error if both `if (!f)` are true. So I guess we need to include a `return` in the second `if` after the of the `pv_log` call

PR #468, opened 2025-01-16
refactor: remount lxc hook in C

anibalportero:
This PR substitutes the remount.sh script, configured to be run by LXC with the `lxc.hook.mount` key, by a C equivalent binary. Ultimately avoiding time consuming script parsing and directly calling the mount syscall.

In this first version, as with the sh script, we ignore Pantavisor configuration and work directly with fixed paths (`/storage/trails`) and envs (`PV_REMOUNT_POLICY`).

The flow follow these steps:

1. Load and parse the remount directives from `device.json` and `<platform>/src.json` in `/storage/trails/current/.pvr/json`
2. Load and parse all mounted destinations paths and flags from `/proc/mount`
3. Iterate over all mounted paths, applying the remount directives over its current paths


remount/remount.c
```diff
@@ -0,0 +1,605 @@
...
+	if (!(k + 1)) {
+		printf("WARN: value not found\n\r");
+		goto out;
+	}
+
+	struct remount_entry *p;
+	p = calloc(1, sizeof(struct remount_entry));
+	if (!p)
+		goto out;
+	dl_list_add_tail(remount_entries, &p->list);
```
fortizc, line 70: Maybe I'm wrong, but could be an error to add this to the list in this part of the code. Down below, I see a couple of `calloc()` calls not checked, and if they fail, probably you will end with a list with some invalid members?

remount/remount.c
```diff
@@ -0,0 +1,605 @@
...
+#include "utils/json.h"
+#include "utils/str.h"
+
+struct remount_entry {
+	char *exp;
+	char *opts;
+	struct dl_list list; // remount_entry
+};
+
+void parse_remount_entry(char *json, struct dl_list *remount_entries)
```
fortizc, line 43: This function and the others in this file should be `static`

remount/remount.c
```diff
@@ -0,0 +1,605 @@
...
+	pname = getenv("LXC_NAME");
+	if (!pname) {
+		printf("ERROR: LXC_NAME env not set\n\r");
+		return NULL;
+	}
+	printf("DEBUG: container name '%s' to be used\n\r", pname);
+
+	char *path_fmt = "%s/run.json";
+	size_t len = strlen(path_fmt) + strlen(pname) + 1;
+	char *path = calloc(1, len);
```
fortizc, line 216: This is wrong, should be 
```c
calloc(len, sizeof(char))
```

remount/remount.c
```diff
@@ -0,0 +1,605 @@
...
+	char *pname = NULL;
+	pname = getenv("LXC_NAME");
+	if (!pname) {
+		printf("ERROR: LXC_NAME env not set\n\r");
+		return NULL;
+	}
+	printf("DEBUG: container name '%s' to be used\n\r", pname);
+
+	char *path_fmt = "%s/run.json";
+	size_t len = strlen(path_fmt) + strlen(pname) + 1;
```
fortizc, line 215: This is not the right size because you are adding `path_fmt + pname + 1` and the string in `pname` replaces 2 chars in `path_fmt` should be `path_fmt - 2 + pname + 1`, but here is the example where two `snprintf` are needed. Other option is to use a fixed buffer with size `PATH_MAX` and `strdup()`.
I don't know if this could be a source of problems in the future...

remount/remount.c
```diff
@@ -0,0 +1,605 @@
...
+out:
+	if (rel_path)
+		free(rel_path);
+	if (tokv)
+		free(tokv);
+
+	return ret;
+}
+
+#define STORAGE_TRAILS_CURRENT_PVR_JSON_PATH "/storage/trails/current/.pvr/json"
```
fortizc, line 271: Should this type of constants be placed at the beginning of the file?

remount/remount.c
anibalportero: moved that to the end and added checks and freeup for calloc failures

remount/remount.c
anibalportero: done

remount/remount.c
anibalportero: done

remount/remount.c
anibalportero: I think now it's ok

remount/remount.c
anibalportero: done

PR #480, opened 2025-02-20
avoid running export.sh hook when defined in run.json

fortizc:
The field `export` (bool) is added to pv_platform to avoid run the export hook when the platform has not a export definition.
* the field is initialized as false (platforms.c)
* the field is set to true during the state json parsing. (parser_system1.c)
* export hook is not added if the export is false. (pv_lxc.c)

### DIFF
--- a/parser/parser_system1.c
+++ b/parser/parser_system1.c
@@ -1221,6 +1221,14 @@ static int do_action_for_drivers(struct json_key_action *jka, char *value)
 	return 0;
 }
 
+static int do_action_for_export(struct json_key_action *jka, char *value)
+{
+	struct platform_bundle *bundle = (struct platform_bundle *)jka->opaque;
+	struct pv_platform *platform = *bundle->platform;
+	platform->export = true;
+	return 0;
+}
+
 static int parse_platform(struct pv_state *s, char *buf, int n)
 {
 	char *config = NULL, *shares = NULL;
@@ -1262,6 +1270,8 @@ static int parse_platform(struct pv_state *s, char *buf, int n)
 			      do_action_for_storage, false),
 		ADD_JKA_ENTRY("drivers", JSMN_OBJECT, &bundle,
 			      do_action_for_drivers, false),
+		ADD_JKA_ENTRY("exports", JSMN_ARRAY, &bundle,
+			      do_action_for_export, false),
 		ADD_JKA_NULL_ENTRY()
 	};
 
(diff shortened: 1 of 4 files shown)


parser/parser_system1.c
```diff
@@ -1262,6 +1270,8 @@ static int parse_platform(struct pv_state *s, char *buf, int n)
 			      do_action_for_storage, false),
 		ADD_JKA_ENTRY("drivers", JSMN_OBJECT, &bundle,
 			      do_action_for_drivers, false),
+		ADD_JKA_ENTRY("exports", JSMN_ARRAY, &bundle,
```
anibalportero, line 1273: Is what we expect an array and not a string?

anibalportero (commented):
As with previous tasks involving not documented features, it would be nice if you could take some time into updating what is expected from run.json syntax in https://docs.pantahub.com/pantavisor-state-format-v2/#containerrunjson 

Not not what this is doing, but a small explanation of what is the expected behavior for this feature from a user persective would be nice in https://docs.pantahub.com/containers/#storage

parser/parser_system1.c
fortizc: I guess... as you notice we lack of docs about it, my only sources of docs are 

- https://github.com/pantavisor/pantavisor/blob/master/scripts/hooks_lxc-mount.d/export.sh
- https://gitlab.com/pantacor/pvr/-/blob/master/templates/builtin-lxc-docker.go?ref_type=heads

fortizc:
> As with previous tasks involving not documented features, it would be nice if you could take some time into updating what is expected from run.json syntax in https://docs.pantahub.com/pantavisor-state-format-v2/#containerrunjson
> 
> Not not what this is doing, but a small explanation of what is the expected behavior for this feature from a user persective would be nice in https://docs.pantahub.com/containers/#storage

That the thing, this is not a new feature in terms of what the user see... 
Until now the `export.sh`  always run but just until the point where it checks if the "exports" keys exists in the json. Now that check is made at the moment of the state json parse and the hook is not added if isn't needed. For that reason is a fix (see the branch name) because improve or fix the current behavior

anibalportero:
I can see it's not documented and I can see it's not a new feature. I'm just saying it would be nice to understand a bit what we are changing here. Besides that, if we can take the chance and document it, so in the future we don't find ourselves in the same problem when we need to touch this again, that would be great.

PR #481, opened 2025-02-21
fix: update status and retry count not being loaded after processing new updates and rolling back

anibalportero:
This PR fixes Pantavisor not having into account hub update status of stored revisions when retrying them. It does the same for stored revisions after a rollback.

It is intended to fix several scenarios:

* Unexpected rollbacks, caused by crashes or power-cycles, were not being set as ERROR neither in storage nor hub.
* Revisions in hub could be stuck in TESTING and Pantavisor would not retry or ERROR them.
* When processing a remote update, we were not having into account the status it might have in disk from previous attempts. This might lead to Pantavisor retrying a revision that had been set as ERROR in disk but not in Hub.
* Retry number was being kept in the cloud, leading to potential unnecessary retries in case the messages would not reach the hub.

It contains this list of changes:

* When booting up and coming from a rollback, Pantavisor now tries to load the existing progress from disk. In case the progress found is not final (ERROR, WONTGO, UPDATED or DONE), we set it both in storage and hub as ERROR with message "Unexpected rollback".
* We now process TESTING along the rest of pending revision status (NEW, QUEUED, DOWNLOADING and INPROGRESS) at each idle client step.
* When processing a pending revision, we first try to load its existing progress, if any, from disk. If the progress status exists and is final, we stop processing that revision and try to set that value from storage into hub.
* Retry number is now stored in disk, so we don't rely on requests that might not reach the cloud. Furthermore, its scope has been amplified from DOWNLOADING only to include any pending revision processing (NEW, QUEUED, INPROGRESS and TESTING). This means, Pantavisor will not try to process an update unlimited times and this protects us from possible loops and unnecessary retries.


asac:
if i understand this patch would make the device side the one with the final say about what is going on. This is good. But, is there a chance for hub - if reachable - to suggest things to pantavisor still? e.g. like cancel more retries etc?

anibalportero:
> if i understand this patch would make the device side the one with the final say about what is going on. This is good. But, is there a chance for hub - if reachable - to suggest things to pantavisor still? e.g. like cancel more retries etc?

Yes, Pantavisor will ignore any revision that is not in the pending list (NEW, QUEUED, DOWNLOADING, TESTING and INPROGRESS), so if a revision is set as "WONTGO" or any other string such as "CANCELLED", it will not process that.

PR #482, opened 2025-02-27
Remove broken libthttp log implementation from ph_logger

fortizc:
Remove the libthttp implementation in ph_logger because the current implementation in log.c is working well, so this is broken and duplicated.

Example:
```
[pantavisor] 368 DEBUG	 -- [ph_logger]: getaddrinfo: ࡐ??? (x?????)
[pantavisor] 368 WARN	 -- [ph_logger]: HTTP request POST /logs/ could not auth (status=3)
[pantavisor] 370 DEBUG	 -- [libthttp]: getaddrinfo: Resource temporarily unavailable (Try again)
```
In the example, the first and the third messages are the same, but the first has the problem. This is because `libthttp` send the logs in "args" so call `va_star` and ph_logger use `vlog()` which calls `va_start` again and that an UB

### DIFF
--- a/ph_logger.c
+++ b/ph_logger.c
@@ -394,7 +394,7 @@ static int ph_logger_push_from_file(const char *filename, char *platform,
 			int len = newline_at - src + 1;
 			/*
 			 * Use json_holder temporarily to
-			 * get the source name and platform 
+			 * get the source name and platform
 			 * name.
 			 */
 			SNPRINTF_WTRUNC(json_holder, len, "%.*s", len - 1, src);
@@ -643,14 +643,6 @@ static int ph_logger_push_revision(char *revision)
 	return result;
 }
 
-static void log_libthttp(int level, const char *fmt, va_list args)
-{
-	if (level > pv_config_get_int(PV_LIBTHTTP_LOG_LEVEL))
-		return;
-
-	vlog(MODULE_NAME, DEBUG, fmt, args);
-}
-
 static pid_t ph_logger_start_push_service(char *revision)
 {
 	pid_t helper_pid = -1;
@@ -678,7 +670,6 @@ static pid_t ph_logger_start_push_service(char *revision)
 		       "Initialized push service with pid %d by process with pid %d",
 		       getpid(), getppid());
 		pv_log(DEBUG, "Push service pushing logs for rev %s", revision);
-		thttp_set_log_func(log_libthttp);
 		while (1) {
 			// if nothing to push or error while pushing, sleep
 			if (ph_logger_push_revision(revision) <= 0) {
@@ -778,7 +769,6 @@ static pid_t ph_logger_start_range_service(struct pantavisor *pv,
 		pv_log(INFO,
 		       "Initialized range service with pid %d by process with pid %d",
 		       getpid(), getppid());
-		thttp_set_log_func(log_libthttp);
 		while (current_rev >= 0) {
 			// skip current revision.
 			if (atoi(avoid_rev) == current_rev) {


anibalportero:
I don't think this solution is good. In this case, the messages might be the same, but they are logging different HTTP requests.

There is one call to `thttp_set_log_func` for each fork that is using libthttp. That way, the requests that are made in each fork can be logged. In this case, all requests are failing because the server is unreachable, so everyone is complaining about the same problem. It could be that the requests that are made by each fork could get different logging, because of bad request format, bad endpoint and so on.

fortizc:
> I don't think this solution is good. In this case, the messages might be the same, but they are logging different HTTP requests.
> 
> There is one call to `thttp_set_log_func` for each fork that is using libthttp. That way, the requests that are made in each fork can be logged. In this case, all requests are failing because the server is unreachable, so everyone is complaining about the same problem. It could be that the requests that are made by each fork could get different logging, because of bad request format, bad endpoint and so on.

Are you sure? because each fork inherits the memory of the main process, so

fortizc:
@anibalportero I think this solution is better, it works just like normal, but call logserver directly to avoid the double `va_start`

ph_logger.c
```diff
@@ -689,6 +693,8 @@ static pid_t ph_logger_start_push_service(char *revision)
 		       "Initialized push service with pid %d by process with pid %d",
 		       getpid(), getppid());
 		pv_log(DEBUG, "Push service pushing logs for rev %s", revision);
+
+		current_process = "push_service_libthttp";
```
anibalportero, line 697: this is a good idea. Can you change the other one in log.c to something like "ph_client_libthttp" so they are all using the same format?

anibalportero (commented):
I suggest we start using some naming convention for commits. For example, starting with "fix:", "feature:", "refactor:"... And then adding some clear and explicit description about what is this fixing.

In this case, I'd put something similar to "fix: libthttp logs inside ph_logger forks writing garbage".

nandra:
> I suggest we start using some naming convention for commits. For example, starting with "fix:", "feature:", "refactor:"... And then adding some clear and explicit description about what is this fixing.
I'm using this in some projects: https://www.conventionalcommits.org/en/v1.0.0/ and it's then nice to generate changelog also ;)
> 
> In this case, I'd put something similar to "fix: libthttp logs inside ph_logger forks writing garbage".

anibalportero:
> > I suggest we start using some naming convention for commits. For example, starting with "fix:", "feature:", "refactor:"... And then adding some clear and explicit description about what is this fixing.
> > I'm using this in some projects: https://www.conventionalcommits.org/en/v1.0.0/ and it's then nice to generate changelog also ;)
> > In this case, I'd put something similar to "fix: libthttp logs inside ph_logger forks writing garbage".

That looks good. I'll add that to a new CONTRIBUTING.md :)

fortizc:
> > > I suggest we start using some naming convention for commits. For example, starting with "fix:", "feature:", "refactor:"... And then adding some clear and explicit description about what is this fixing.
> > > I'm using this in some projects: https://www.conventionalcommits.org/en/v1.0.0/ and it's then nice to generate changelog also ;)
> > > In this case, I'd put something similar to "fix: libthttp logs inside ph_logger forks writing garbage".
> 
> That looks good. I'll add that to a new CONTRIBUTING.md :)
 
This is great, we need some guidelines to avoid ambiguity

ph_logger.c
fortizc: Sure, but are you sure that there are no others modules which are using that log?

anibalportero: Yes, that is for all the requests that we are doing from the main fork. "ph_client_libthttp" its not very specific though and will fit well with almost anything.

fortizc: Done

PR #484, opened 2025-03-05
Fix interger overflows

fortizc:
Fix some possible integer overflows in memory assigns, and maybe the most relevant in the libthttp.

**IMPORTANT:** to merge this PR, please fist check and merge this: https://gitlab.com/pantacor/libthttp/-/merge_requests/49 which implements the solution in the libthttp side


anibalportero:
In this case, I'd put "fix: libthttp size offset overflow and several bad memory allocations"

fortizc:
> In this case, I'd put "fix: libthttp size offset overflow and several bad memory allocations"

So let me change it using the standard propsed by Marek

PR #485, opened 2025-03-05
fix: ctrl continues replying after OK failed because of broken pipe

anibalportero:
This only applies to GET requests in ctrl. In those cases, we send an OK after receiving the request and before replying. Before the patch, if  the OK would fail because of broken pipe, we would resume replying. Now, we stop doing that so we don't write on a closed fd.


anibalportero:
broken pipe signal ignored in https://github.com/pantavisor/pantavisor/pull/489

PR #490, opened 2025-03-07
fix: avoid checksum calculation over the same file

fortizc:
Create an array to store the inode of each file with an already calculated checksum to avoid the recalculation

### DIFF
--- a/storage.c
+++ b/storage.c
@@ -466,31 +466,21 @@ out:
 	return ret;
 }
 
-static bool pv_storage_validate_objects_object_checksum(char *checksum)
-{
-	char path[PATH_MAX];
-
-	pv_paths_storage_object(path, PATH_MAX, checksum);
-	pv_log(DEBUG, "validating checksum for object %s", path);
-	return !pv_storage_validate_file_checksum(path, checksum);
-}
-
 bool pv_storage_validate_trails_object_checksum(const char *rev,
 						const char *name,
 						char *checksum)
 {
-	char path[PATH_MAX];
+	char trail[PATH_MAX] = { 0 };
+	char object[PATH_MAX] = { 0 };
 
-	// validate object in pool to match
-	if (!pv_storage_validate_objects_object_checksum(checksum)) {
-		pv_log(ERROR, "object %s with checksum %s failed", name,
-		       checksum);
+	pv_paths_storage_trail_file(trail, PATH_MAX, rev, name);
+	pv_paths_storage_object(object, PATH_MAX, checksum);
+
+	if (!pv_fs_file_is_same(trail, object))
 		return false;
-	}
 
-	pv_paths_storage_trail_file(path, PATH_MAX, rev, name);
-	pv_log(DEBUG, "validating checksum for object in trail %s", path);
-	return !pv_storage_validate_file_checksum(path, checksum);
+	pv_log(DEBUG, "validating checksum for %s and %s", object, trail);
+	return !pv_storage_validate_file_checksum(trail, checksum);
 }
 
 bool pv_storage_validate_trails_json_value(const char *rev, const char *name,
--- a/utils/fs.c
+++ b/utils/fs.c
@@ -432,3 +432,27 @@ int pv_fs_file_check_and_open(const char *fname, int flags, mode_t mode)
 		return -1;
 	return open(fname, flags, mode);
 }
+
+static int pv_fs_file_inode_get(const char *path, ino_t *inode)
+{
+	struct stat st = { 0 };
+	if (stat(path, &st) != 0)
+		return -1;
+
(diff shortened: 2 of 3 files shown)


anibalportero:
I'm not sure about this change. Maybe I'm wrong, but to me it feels completely off.

First, the double checksum validation is done twice per bootup. DONE should only be set once per update, so the free up will not cover all boot ups?

Also, I don't know why do we have to use a global variable. The objects already have their info stored in pv_state in the pv_object struct.

Finally, the list inode approach doesn't seem like the most efficient. You have the double checksum in pv_storage_validate_trails_object_checksum. Wouldn't be easier to just compare the inode of the two files in there, where you already have the two paths that are being validated?

fortizc:
> I'm not sure about this change. Maybe I'm wrong, but to me it feels completely off.
> 
> First, the double checksum validation is done twice per bootup. DONE should only be set once per update, so the free up will not cover all boot ups?

Ok, got it you are right.

> Also, I don't know why do we have to use a global variable. The objects already have their info stored in pv_state in the pv_object struct.
> Finally, the list inode approach doesn't seem like the most efficient. You have the double checksum in pv_storage_validate_trails_object_checksum. Wouldn't be easier to just compare the inode of the two files in there, where you already have the two paths that are being validated?

The thing is, that solution has the problem that you would need to loop through the whole `pv_object` list looking for the path. That is the most space efficient, but not the most time efficient

fortizc:
> Finally, the list inode approach doesn't seem like the most efficient. You have the double checksum in pv_storage_validate_trails_object_checksum. Wouldn't be easier to just compare the inode of the two files in there, where you already have the two paths that are being validated?

But this means that I should look using path and/or checksum in pv_object (objpath/relpath) and if I found it I should compare the inodes... then how I know if the checksum was calculated? I should add a new boolean field?  and more important I should check the object in `pv->state->object` or `pv->update->state->object`

anibalportero:
> > Finally, the list inode approach doesn't seem like the most efficient. You have the double checksum in pv_storage_validate_trails_object_checksum. Wouldn't be easier to just compare the inode of the two files in there, where you already have the two paths that are being validated?
> 
> But this means that I should look using path and/or checksum in pv_object (objpath/relpath) and if I found it I should compare the inodes... then how I know if the checksum was calculated? I should add a new boolean field? and more important I should check the object in `pv->state->object` or `pv->update->state->object`

I just think iterate twice is not needed, as pv_object already has both enough info to be able to check both paths at the same time. 

In pv_state_validate_checksum, we iterate over all objects, then in pv_storage_validate_trails_object_checksum we validate them both. Isn't that the double validation we want to remove? Wouldn't it be possible to just get the inode and validate checksum of the first, then get inode from the second, compare the two and if checksum  of the first and both inode are the same we return true? It feels to me that everything could be done inside pv_storage_validate_trails_object_checksum without having to iterate again and maybe even without having to store inode in pv_object.

storage.c
```diff
@@ -466,31 +466,58 @@ int pv_storage_validate_file_checksum(char *path, char *checksum)
...
+	return true;
+}
+
+
 bool pv_storage_validate_trails_object_checksum(const char *rev,
 						const char *name,
 						char *checksum)
 {
-	char path[PATH_MAX];
+	char file[PATH_MAX] = { 0 };
```
anibalportero, line 510: If the other one is object, I'd call this one trail. Both are files.

storage.c
```diff
@@ -466,31 +466,58 @@ int pv_storage_validate_file_checksum(char *path, char *checksum)
...
+	}
 
-	pv_paths_storage_object(path, PATH_MAX, checksum);
-	pv_log(DEBUG, "validating checksum for object %s", path);
-	return !pv_storage_validate_file_checksum(path, checksum);
+	*inode = st.st_ino;
+	return 0;
 }
 
+static bool pv_storage_compare_files(const char *object, const char *file)
```
anibalportero, line 483: This function is comparing two generic files, so I'd call it something like is_same_file(path1, path2) and maybe move it to utils/fs.c?

storage.c
```diff
@@ -466,31 +466,58 @@ int pv_storage_validate_file_checksum(char *path, char *checksum)
...
+	pv_log(DEBUG, "validating inodes for %s and %s", object, file);
+
+	if (pv_storage_get_inode(object, &obj_ino) != 0)
+		return false;
+	if (pv_storage_get_inode(file, &file_ino) != 0)
+		return false;
+
+	if (obj_ino != file_ino) {
+		pv_log(DEBUG,
+		       "inode mismatch! can not validate checksum (obj: %s file: %s)",
```
anibalportero, line 497: you would have to move this outside in pv_storage_compare_files

PR #498, opened 2025-04-11
updater.c: fix revision check not reporting stale if new_rev == current_rev

asac:


updater.c
```diff
@@ -872,7 +872,7 @@ static int trail_get_new_steps(struct pantavisor *pv)
 	if (!errno && (new_rev <= current_rev)) {
 		pv_log(WARN, "running revision %ld but trying to update to %ld",
 		       current_rev, new_rev);
-		if (new_rev < current_rev) {
+		if (new_rev <= current_rev) {
```
anibalportero, line 875: I think this should remain <, as we now load whatever is in /storage/trails/current/.pv/progress (updater.c:886). Then, if we find a final state (updater.c:903), we try to report that back, which is probably better than a generic "stale" error.

anibalportero: This should not be the case since https://github.com/pantavisor/pantavisor/pull/481. The idea now is to report to hub whatever is on disk if we find something final (DONE, UPDATED, WONTGO or ERROR).

asac:
after discussion i feel that new code should not cause these issues anymore....

PR #499, opened 2025-04-14
tsh: add tsh_run_logserver wrapper

asac:
We register the read side of the tsh io pipe with logserver


anibalportero:
We are trying to follow a commit message format to make release changelogs nicer. Would be good if you can change that :)

https://docs.pantahub.com/pantavisor-contrib-guidelines/

utils/tsh.c
```diff
@@ -193,6 +201,61 @@ pid_t tsh_run_io(char *cmd, int wait, int *status, int stdin_p[],
...
+	int ret = 0;
+	int out_pipe[2] = { 0 };
+	int err_pipe[2] = { 0 };
+
+	if (logserver_subscribe_pipe(out_pipe, log_source_out, INFO) != 0 ||
+	    logserver_subscribe_pipe(err_pipe, log_source_err, WARN) != 0) {
+		return -1;
+	}
+
+	ret = tsh_run_io(cmd, 1, wstatus, NULL, out_pipe, err_pipe);
```
fortizc, line 230: Here I saw something wrong, the third parameter of `tsh_run_io` (wstatus) should be a pointer to that variable

asac:
> We are trying to follow a commit message format to make release changelogs nicer. Would be good if you can change that :)
> 
> https://docs.pantahub.com/pantavisor-contrib-guidelines/

ok done that.

PR #500, opened 2025-04-15
perf: introduce event loop

anibalportero:
This is the first delivery of a series of PRs that aim at eliminating all long blocking routines in the `Pantavisor` main loop.  This will ultimately allow faster `pvcontrol` response times, reduce boot up time, increase pvcontrol availability, have a reliable watchdog always mode and start using a better tested HTTP client and server. To achieve all that, it starts with the transformation of the main loop into an event loop, with the help of the [libevent library](https://github.com/libevent/libevent).

The first PR of this series started just including the conversion of the user metadata `GET` and device metadata `PUT` requests into non-blocking events, using callbacks to process each of the responses, also paving the way to move all client code into `pantahub/`, including its state machine.

In order to make the non-blocking metadata exchange with Hub work within an acceptable time, we also had to move the `Pantavisor` state machine into events themselves, as well as removing the existing 2 second select for `pv-ctrl`. This is done to make room for `libevent` to process the internal events generated by HTTP `libevent` (several per request).

In summary, the events that have been added in this first delivery are:

- Periodic **devmeta upload**.
- Periodic **usrmeta download**.
- **Devmeta upload response** callback.
- **Usrmeta download response** callback.
- One shot **states** in the **pantavisor state machine**, minus WAIT.
- **Periodic WAIT** from the pantavisor state machine.
- **pv-ctrl socket select**.

For fine tuning the utilization of libevent, these config keys have been added:

- `PH_LIBEVENT_HTTP_TIMEOUT`: sets the timeout for libevent HTTP requests, including connect, read and write only. Therefore, excluding time dedicated to other events.
- `PH_LIBEVENT_HTTP_RETRIES`: sets the timeout for libevent HTTP retries in case of failure.
- `PV_LIBEVENT_DEBUG_MODE`: activates libevent debug mode and verbose logs.

Further deliveries:

- `LXC` plugin container start.
- `pv-ctrl` replies.
- Hub updates.
- Hub claim.
- `wdt` always.
- Increase `pv-ctrl` availability.
- Remove libthttp.

About using `PUT` instead of `PATCH` for device metadata: Using `PUT` will result in requests with bigger lengths. On the other hand, we will gain the ability to remove device metadata from cloud and get a simpler code in general. These request will still be performed only when there is new device metadata to upload.


pantahub/pantahub_proto.c
```diff
@@ -0,0 +1,211 @@
...
+{
+	return cache && in && pv_str_matches(cache, strlen(cache), in, strlen(in));
+}
+
+static void _save_cache(const char *in, char **cache)
+{
+	if (*cache)
+		free(*cache);
+	*cache = strdup(in);
+}
```
asac, line 139: this cache concept is not clear to me.... what is it and should we really have this globally rater than inside pv?

pantahub/pantahub_rest.c
```diff
@@ -0,0 +1,301 @@
...
+
+	while (n--) {
+		if (!strncmp(files[n]->d_name, ".", 1)) {
+			free(files[n]);
+			continue;
+		}
+
+		pv_paths_cert(path, PATH_MAX, files[n]->d_name);
+		size = strlen(path);
+		cafiles[i] = malloc((size + 1) * sizeof(char));
```
fortizc, line 112: here is better to call `calloc()` to avoid a possible integer overflow

pantahub/pantahub_proto.c
anibalportero: I ended up hanging it from a new pantahub_t global similar to pv struct. I removed user meta cache, which is redundant, and left devmeta cache so we know when to PUT. Let me know if you don't like it.

pantahub/pantahub_proto.c
```diff
@@ -0,0 +1,211 @@
...
+		 pv_config_get_str(PH_CREDS_ID));
+
+	pv_pantahub_rest_send(base, EVHTTP_REQ_GET, uri, session.token, NULL, _recv_get_usrmeta_cb);
+}
+
+static void _recv_set_devmeta_cb(struct evhttp_request *req, void *ctx)
+{
+	char buffer[1024];
+	memset(&buffer, 0, sizeof(buffer));
+	if (pv_pantahub_rest_recv(req, ctx, &buffer[0], 1024) != 200) {
```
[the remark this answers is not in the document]
anibalportero: done

pantahub/pantahub_rest.c
```diff
@@ -0,0 +1,301 @@
...
+		_add_header(output_headers, "Content-Type", "application/json");
+	}
+
+	res = evhttp_make_request(ctx->evcon, req, op, uri);
+	if (res != 0) {
+		pv_log(ERROR, "evhttp_make_request returned code %d", res);
+		goto error;
+	}
+
+	return 0;
```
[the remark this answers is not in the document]
anibalportero: ctx is passed as an argument to the callback by evhttp_request_new. Then, it is freed by the library, be it after a successful reply or in evhttp_connection_free in case of error. It's a bit confusing but that's how libevent manages it.

PR #501, opened 2025-04-16
fix: env conf not overwriting ph.conf and OEM conf

anibalportero:
Before this PR, configuration level overwrite would be defined by its loading order (defaults -> pv.conf -> pv_ cmdline -> env -> policy -> OEM -> ph.conf -> ph_ cmdline -> user-meta -> command). 

This would present two problems:

* Env would not overwrite policy and OEM configs.
* Legacy pv_ cmdline would not overwrite policy and OEM configs.

This fix, preserves the loading order of the different levels, as that will allow to load PV_POLICY and PV_OEM_NAME in the right places, while it will introduce a new hierarchy of levels. The new hierarchy prevents lower levels to overwrite upper levels, following this order:

defaults < pv.conf/ph.conf < policy < OEM < cmdline < env < user-meta < command

Besides the parsing order, the change also keeps the allowed levels for each key untouched.

### DIFF
--- a/config.c
+++ b/config.c
@@ -67,10 +67,10 @@ typedef enum {
 	PH_CONF = 1 << 2,
 	PH_CLIENT = 1 << 3,
 	POLICY = 1 << 4,
-	PV_CMDLINE = 1 << 5,
-	PH_CMDLINE = 1 << 6,
-	ENV = 1 << 7,
-	OEM = 1 << 8,
+	OEM = 1 << 5,
+	PV_CMDLINE = 1 << 6,
+	PH_CMDLINE = 1 << 7,
+	ENV = 1 << 8,
 	META = 1 << 9,
 	CMD = 1 << 10
 } level_t;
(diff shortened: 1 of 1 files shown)


config.c
```diff
@@ -788,6 +788,14 @@ static int _set_config_by_entry(struct pv_config_entry *entry,
...
 	}
 
+	if (modified < entry->modified) {
+		pv_log(WARN,
+		       "key '%s' in level %s already configured in upper level %s",
+		       entry->key, _get_mod_level_str(modified),
+		       _get_mod_level_str(entry->modified));
+		return 0;
+	}
+
```
asac, line 798: Not sure if i read this right, but if this means that a lower level config cannot be overwritten in a higher level this would be against the whole spirit of the config layers. Those are there to allow exactly that: refinement through downstream or runtime actions.

PV_POLICY and PV_OEM_NAME should not be configurable on the oem config level. and wont have an effect any other layer higher than CMDLINE.  Cant we configure the configs to not be configurable in those layers in between?

anibalportero: This doesn't change that. Each config entry (see struct pv_config_entry in config.c), has both an allowed and a modified mask with the level_t enum type. These changes only affect the second one.

The allowed mask stores the level in which we can change the config key. It is set during initialization and left unchanged. Then, it is checked a bit before this, in line 785, so we can filter out set attempts in forbidden levels for each key. That is left untouched.

This affects the modified mask though. We used that mask for information purposes only and displayed that with pvcontrol conf ls. Now, besides that, we also use it to check whether a key was previously set from a higher level (taking the level order from the level_t enum). However, this is after the allowed check.

In summary, this would prevent PH_CREDS_HOST being set in env and overwritten in ph.conf, because PH_CONF < ENV, but it would not allow it being set from user meta, because the allow levels are set in a different variable and goes in a different way.

anibalportero: This is the docs commit https://gitlab.com/pantacor/docs/-/commit/9907430465993bff41bfe50718ee628242e02abb.

I think it separates the three main concepts here: when is each level parsed, what are the allowed keys for each level and who can overwrite who.

PR #507, opened 2025-05-20
Native PVTX implementation

fortizc:
This PR adds PVTX as a native C application inside Pantavisor
All the code is inside the `pvtx/` folder and the files are:
* pvtx: this the main application.
* pvtx_ctrl: implements the pv-ctrl communication (http over unix socket).
* pvtx_jsmn_utils: helpers to parse json.
* pvtx_state: state json management.
* pvtx_txn: implements the transactions and queue operations
* pvtx_tar: tar implementation to unpack Pantavisor packages. The following files are part of this lib too:
    * pvtx_tar_impl: defines the interface to implement decompress
    * pvtx_tar_formats: contains the implementation to access to different compressed tar files: gz, bz2 and raw tar (uncompressed)

A new target was added in the main CMakeLists.txt and the installation occurs under /lib/pv/pvtx. 

**NOTE 1:** is order to the `begin` (without arguments) command works, we need to fix/patch the pantavisor web server to include the **content-lenght** on his responses see #506 

**NOTE 2:** currently pvtx work without problems in most of the containers, but in some cases cannot find `libgcc_so.1` so you always can include `target_link_options(pvtx PRIVATE -static-libgcc)` to build statically or include the right lib in the container


CMakeLists.txt
```diff
@@ -240,6 +240,33 @@ target_include_directories(remount PRIVATE ${CMAKE_SOURCE_DIR} ${CMAKE_SOURCE_DI
...
+			pvtx/pvtx_tar.h
+			pvtx/pvtx_tar_impl.h
+			pvtx/pvtx_txn.c
+			pvtx/pvtx_txn.h
+			pvtx/pvtx.c
+)
+target_compile_definitions(pvtx PUBLIC -DDISABLE_LOGSERVER)
+target_include_directories(pvtx PRIVATE ${CMAKE_SOURCE_DIR} ${CMAKE_SOURCE_DIR}/utils)
+target_link_libraries(pvtx ${THTTP} ${MBEDCRYPTO_LIBRARY} ${ZLIB})
+install(TARGETS pvtx DESTINATION /lib/pv)
```
asac, line 267: why dont we install pvtx in normal bindir?

pvtx/pvtx.c
```diff
@@ -0,0 +1,318 @@
...
+
+	int ret = pv_pvtx_queue_process(var[0], var[1], var[2], &err);
+
+	if (ret != 0)
+		fprintf(stderr, "ERROR: %s\n", err.str);
+
+	return ret;
+}
+
+static int pv_pvtx_proccess_args(int argc, char **argv)
```
asac, line 286: lets fix typos: proccess == process

pvtx/pvtx_tar_formats.c
```diff
@@ -0,0 +1,263 @@
...
+// The idea is to track the bytes read (see the read function) and to move it
+// we use a read operation.
+static off_t pv_pvtx_bzip2_seek(void *impl_data, off_t offset, int whence)
+{
+	struct pvtx_bz2 *bz2 = (struct pvtx_bz2 *)impl_data;
+
+	off_t to_move = offset;
+	if (whence == SEEK_SET) {
+		if (offset < bz2->pos) {
+			printf("1===\n");
```
asac, line 198: debug printfs

CMakeLists.txt
fortizc: I just thought that was the right place (I saw the remount example in the same file), so move it to /usr/bin would be ok ?

pvtx/pvtx_tar_formats.c
fortizc: 🤦🏼

pvtx/pvtx.c
fortizc: 🤦🏼

PR #510, opened 2025-05-26
refactor debug experience

ferlzc:
To sum up this PR:

- Adds the option to access the debug shell at any time by pressing <Enter> on the serial console. Previously, the debug shell was only accessible during system initialization.
- You can now exit the debug shell by typing exit, and re-enter it by pressing <Enter> again.
- If a debug shell session is active and the system is preparing to reboot for any reason, the reboot is deferred for 60 seconds. A timeout is triggered during this period.
- A warning message is sent to the active shell indicating the impending reboot.
- During the timeout phase, you can further defer the reboot by using pvcontrol CMD_DEFER_REBOOT <value> to set a new timeout.
- When the timeout reaches 10 seconds, a final warning message is issued to the active debug shell.
- Once the timeout completes, the system proceeds with the reboot.


pantavisor.c
```diff
@@ -622,6 +622,12 @@ static pv_state_t _pv_wait(struct pantavisor *pv)
...
 
+	// check state of debug shell
+    if (pv_debug_check_shell_running()) {
+        next_state = PV_STATE_POWEROFF;
+        goto out;
+    }
+
 	// receive new command. Set 2 secs as the select max blocking time, so we can do the
 	// rest of WAIT operations
 	pv->cmd = pv_ctrl_socket_wait(pv->ctrl_fd, 2);
```
ferlzc, line 633: The main idea here is to check on during pv_wait state to see if a debug shell is running.

debug.c
```diff
@@ -72,14 +75,27 @@ void pv_debug_start_shell()
...
+
+		pv_config_set_debug_shell_active(true);
+
+		timer_start(&timer_debug_shell_timeout,
+			    pv_config_get_int(PV_DEBUG_SHELL_TIMEOUT), 0,
+			    RELATIV_TIMER);
+
+		pv_log(DEBUG, "DEBUG SHELL timeout started with %d secs",
+		       PV_DEBUG_SHELL_TIMEOUT);
+	}
```
ferlzc, line 92: This will set that a debug shell is active and start a timeout

debug.c
```diff
@@ -146,6 +162,39 @@ void pv_debug_check_ssh_running()
...
+
+		if (pid == shell_pid) {
+			pv_log(INFO, "DEBUG SHELL (pid=%d) ended with %d",
+			       shell_pid, WEXITSTATUS(status));
+			shell_pid = -1;
+			pv_config_set_debug_shell_active(false);
+            return false;
+		}
+
+		if (timeout_debug_shell.fin) {
```
ferlzc, line 188: In the case that debug shell timetout expires, we will return true to this function in with will trigger a state change on pantavisor.c, moving from pv_wait to pv_poweroff

anibalportero:
This will be a nice addition :)

Remember that we were trying to use a commit name convention https://docs.pantahub.com/pantavisor-contrib-guidelines/.

config.c
```diff
@@ -249,7 +249,9 @@ static struct pv_config_entry entries[] = {
 	{ BOOL, "PV_UPDATER_USE_TMP_OBJECTS", PV | OEM | RUN, 0,
 	  .value.b = false },
 	{ WDT_MODE, "PV_WDT_MODE", PV, 0, false, .value.i = WDT_SHUTDOWN },
-	{ INT, "PV_WDT_TIMEOUT", PV, 0, false, .value.i = 15 }
+	{ INT, "PV_WDT_TIMEOUT", PV, 0, false, .value.i = 15 },
+	{ BOOL, "PV_DEBUG_SHELL_ACTIVE", PV, 0, false, .value.b = false },
```
anibalportero, line 253: please, use alphabetic order

debug.c
```diff
@@ -46,13 +48,58 @@
 
 static pid_t db_pid = -1;
 static pid_t shell_pid = -1;
+static struct timer timer_shell_timeout;
+
+bool pv_parse_positive_long(const char *str, long *out)
```
anibalportero, line 53: might be static and somewhere in utils?

debug.c
```diff
@@ -72,18 +119,74 @@ void pv_debug_start_shell()
...
 			tsh_run("/sbin/getty -n -l /bin/sh 0 console", 0, NULL);
+
+		wall("A reboot timeout is set for 60 seconds. "
+		     "To defer the reboot, run 'pvcontrol defer-reboot [new timeout]'");
+
+		pv_log(INFO, "DEBUG SHELL started with pid %d", shell_pid);
+	}
+}
+
+void pv_debug_stop_shell()
```
anibalportero, line 141: This one could be static too.

debug.c
```diff
@@ -46,44 +48,107 @@
...
+		return;
+
+	sigset_t blocked_sig, old_sigset;
+	int status = 0;
+	sigemptyset(&blocked_sig);
+	sigaddset(&blocked_sig, SIGCHLD);
+	/*
+		 * Block SIGCHLD while we want to wait on this child.
+		 * */
+	sigprocmask(SIG_BLOCK, &blocked_sig, &old_sigset);
```
fortizc, line 89: is this the same that pvsignals_block_chld() does?

wall.c
```diff
@@ -0,0 +1,61 @@
...
+	dprintf(con_fd, "\r\n");
+	va_end(args);
+}
+
+void wall_ssh_users(const char *message)
+{
+	struct utmp *entry;
+	setutent();
+	while ((entry = getutent()) != NULL) {
+		if (entry->ut_type == USER_PROCESS) {
```
fortizc, line 35: here you could change the condition to be negative and `continue` to avoid 1 indentation level

wall.c
```diff
@@ -0,0 +1,61 @@
...
+			char tty_path[128];
```
fortizc, line 36: Why 128? I think this should be PATH_MAX, or maybe I'm wrong?

wall.c
```diff
@@ -0,0 +1,61 @@
...
+			snprintf(tty_path, sizeof(tty_path), "/dev/%s",
+				 entry->ut_line);
+
+			int fd = open(tty_path, O_WRONLY | O_NOCTTY);
+			if (fd >= 0) {
+				dprintf(fd, "%s\r\n", message);
```
fortizc, line 42: Why the `\r\n` ? is this the standard way?

wall.h
```diff
@@ -0,0 +1,8 @@
+#ifndef WALL_H
+#define WALL_H
+
+void wall(const char *message, ...);
+void wall_ssh_users(const char *message);
+void wall_welcome(void);
```
fortizc, line 6: public functions and define guards should be prefixed with `pv_` or `PV_`  as appropriate

debug.c
```diff
@@ -46,44 +48,107 @@
...
+		pv_log(INFO, "DEBUG SHELL timeout reached, rebooting...");
+		return true;
+	}
+}
+
+void pv_debug_defer_reboot_shell(const char *payload)
+{
+	int new_timeout = 0;
+
+	new_timeout = atoi(payload);
```
fortizc, line 147: the use of `atoi()` is discouraged by the man pages, please use something like `strtol()` which allow checking the errors and in this way you can avoid the cast to `long` in the next call to `timer_strat()`

debug.c
```diff
@@ -46,44 +48,107 @@
...
-		return;
-
 	con_fd = open("/dev/console", O_RDWR);
 	if (con_fd < 0) {
 		pv_log(WARN, "Unable to open /dev/console");
 		return;
 	}
 
-	dprintf(con_fd, "Press [d] and [ENTER] for debug ash shell... ");
 	fcntl(con_fd, F_SETFL, fcntl(con_fd, F_GETFL) | O_NONBLOCK);
```
fortizc, line 65: I know that this is not part of your modification, but it would be nice to remove this call in favour of set O_NONBLOCK directly in the `open()` call

debug.c
ferlzc: I didn't know about this function! I've change to use it.

wall.c
ferlzc: No reason to use 128, changed to PATH_MAX.

debug.c
ferlzc: Got it! Much better to use strtol().

debug.c
ferlzc: I've change to non blocking and tested, it was much better to get the shell running.

fortizc: I don't get it, the fd is currently non blocking (from line 65 onwards), my only suggestion is to avoid the `fcntl()` call and do what it does directly in the `open()` call. So you can replace 59 with:
```c
con_fd = open("/dev/console", O_RDWR | O_NONBLOCK);
```
and then delete the `fcntl()` call. Now I wonder if that fd should be O_CLOEXEC too

ferlzc: thanks for noticing that!  about the  `O_CLOEXEC`, I was checking this  [doc](https://man7.org/linux/man-pages/man2/open.2.html#:~:text=O_CLOEXEC%20(since%20Linux%202.6.23))  and it seems appropriate. From my understanding, this `open() `call is not used in a multithreaded environment, because Pantavisor runs in a single process (though I'm not entirely sure about that :shrug:). But since `/dev/console` can be used by another process, I think `O_CLOEXEC` is useful.

asac:
Here what does not work:

1. enter debug shell
2. enter pvr-sdk container
3. send a reboot update to pantavisor

Issues:
1. i dont seem to get the wallop
2. device does not reboot at all; it is looping in these two states

[pantavisor] 561 DEBUG	 -- [controller]: going to state = STATE_DEBUG
[pantavisor] 561 DEBUG	 -- [controller]: going to state = STATE_WAIT

3. if i exit shell i get this error:
[pantavisor] 532 WARN	 -- [ctrl]: could not select ctrl socket with fd 12: Interrupted system call

Also i got the pantavisor banner at some point without me entering or leaving the shell. Felt as if i got that when the reboot should have been triggered after sending the update.

asac:
Also I would lke to see a message on console (or a pantavisor banner) explaining that you can hit ENTER to enter the debug shell.

Also please test the PV_DEBUG_AUTOLOGIN config will make the debug shell enter automatically still.

config.h
```diff
@@ -149,6 +151,7 @@ void pv_config_set_creds_secret(char *secret);
 void pv_config_set_debug_shell(bool shell);
 void pv_config_set_debug_shell_autologin(bool autologin);
 void pv_config_set_debug_ssh(bool ssh);
+void pv_config_set_debug_shell_active(bool shell);
```
anibalportero, line 154: Sorry for not seeing this before. Usually, we store session variables somewhere under the pv struct or as a static, but config are for user set values. There are some exceptions but they have their own defined level, and no level fits with this, which is changed by pantavisor policy. The one that you are using (ARGS) is for arguments that can be set to the pantavisor command (see init.c).

In this case, it would be weird from user perspective if pvcontrol conf ls would return something other than what it was configured by the user.

For me, it is ok to keep maybe some bool PV_DEBUG_SHELL instead of PV_DEBUG_SHELL_ACTIVE, so user can disable/enable this feature. However, session on/off should go in pv struct or somewhere else.

config.h
```diff
@@ -60,6 +60,8 @@ typedef enum {
 	PV_CONTROL_REMOTE_ALWAYS,
 	PV_DEBUG_SHELL,
 	PV_DEBUG_SHELL_AUTOLOGIN,
+	PV_DEBUG_SHELL_ACTIVE,
```
anibalportero, line 63: How is PV_DEBUG_SHELL different from PV_DEBUG_SHELL_ACTIVE. Is it configuring the same shell? is PV_DEBUG_SHELL deprecated now? If possible, please, remove PV_DEBUG_SHELL_ACTIVE and keep PV_DEBUG_SHELL working so we can be backwards compatible with previous configs.

pantavisor.c
```diff
@@ -961,6 +988,9 @@ static pv_state_t pv_shutdown(struct pantavisor *pv, shutdown_type_t t)
 
 static pv_state_t _pv_reboot(struct pantavisor *pv)
 {
+	if (pv_debug_start_timeout_shell()) {
```
anibalportero, line 991: Another question. In this case, we are deferring the reboot, how does PV_STATE_WAIT it has to get back to PV_STATE_REBOOT?

debug.c
ferlzc: Its fixed and now is doing only the `con_fd = open("/dev/console", O_RDONLY | O_NONBLOCK | O_CLOEXEC );
` with O_CLOCKEXE
Removed `fcntl(con_fd, F_SETFL, fcntl(con_fd, F_GETFL) | O_NONBLOCK);` call. I didn't add O_CLOEXEC because I haven't test it. Now is ok. The fcntl I forgot to remove, my mistake = D

debug.c
```diff
@@ -46,44 +51,170 @@
...
-		pv_log(WARN, "waiting for debug shell with pid %d to exit",
-		       shell_pid);
-		waitpid(shell_pid, NULL, 0);
+	int debug_timeout = 0;
+
+	if(shell_timeout_active) {
+		return false;
+	}
+
+	if (pv_config_get_bool(PV_DEBUG_SHELL)) {
```
anibalportero, line 151: I guess you have to guard this with shell_session too? Otherwise, we will enter here as long as PV_DEBUG_SHELL is 1 and even if no shell is opened. Instead of this big indented thingy, I'd do something like:

```
if (!pv_config_get_bool(PV_DEBUG_SHELL) || !shell_session)
    return false;
    
// rest of the stuff
```

debug.c
```diff
@@ -46,44 +51,170 @@
...
+		debug_timeout = pv_config_get_int(PV_DEBUG_SHELL_TIMEOUT);
+		timer_start(&shell_timer, debug_timeout, 0,
+			    RELATIV_TIMER);
+
+		shell_timeout_active = true;
+		pv_log(INFO, "shell timeout started with %d secs",
+		       debug_timeout);
+		pv_wall("System will reboot in 60 seconds... "
```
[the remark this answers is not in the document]
anibalportero: This is a good point. Not sure if it has an easy solution. @fortizc is it easy for him to reuse the pvtx code that makes http requests to pv-ctrl? If not, I would live with it and leave it as technical debt.

A message that could represent better what's to execute right now would be:
```
pventer -c <container-name> pvcontrol defer-reboot [new timeout]
```

And in <container-name>, the user would put pvr-sdk or whatever container that has pvcontrol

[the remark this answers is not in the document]
fortizc: > @fortizc is it easy for him to reuse the pvtx code that makes http requests to pv-ctrl? If not, I would live with it and leave it as technical de

Probably not, for pvtx_ctrl to be able to do this we need to add the request and could be even weird to calla pvtx function for this, maybe when we have the main loop and libevent integrated, it would be easier

wall.c
```diff
@@ -0,0 +1,73 @@
...
+		return;
+	}
+	va_list args;
+	va_start(args, message);
+	vdprintf(con_fd, message, args);
+	dprintf(con_fd, "\n");
+	va_end(args);
+}
+
+void pv_wall_ssh_users(const char *message)
```
fortizc, line 31: this should support variable args to be used like pv_wall

debug.c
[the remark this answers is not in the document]
ferlzc: i'm letting this method for issue a defer-reboot, like this:
pventer -c <container-name> pvcontrol defer-reboot [new timeout]

in the future we can do something as you guys mencioned.

debug.c
```diff
@@ -46,44 +51,170 @@
...
+{
+	struct timer_state state = timer_current_state(&shell_timer);
+
+	if (!state.fin)
+		return state.sec;
+
+	return 0;
+}
+
+bool pv_debug_start_shell()
```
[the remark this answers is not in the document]
ferlzc: still need to ajust this. I will fixed to use int.

pantavisor.c
```diff
@@ -833,6 +854,14 @@ static pv_state_t _pv_rollback(struct pantavisor *pv)
 	return PV_STATE_REBOOT;
 }
 
+static pv_state_t _pv_debug(struct pantavisor *pv)
+{
+	if (pv_debug_check_shell())
```
anibalportero, line 859: Maybe we don't need a state just for this? It looks to me that this could go in 629 and we would get less verbose logs:

```
[pantavisor] 128 DEBUG   -- [controller]: going to state = STATE_DEBUG                                                                                                                                                                      
[pantavisor] 128 DEBUG   -- [controller]: going to state = STATE_WAIT                                                        
[pantavisor] 131 DEBUG   -- [controller]: going to state = STATE_DEBUG                                                                                              
[pantavisor] 131 DEBUG   -- [controller]: going to state = STATE_WAIT                                                       
[pantavisor] 133 DEBUG   -- [controller]: going to state = STATE_DEBUG                  
[pantavisor] 133 DEBUG   -- [controller]: going to state = STATE_WAIT                  
[pantavisor] 135 DEBUG   -- [controller]: going to state = STATE_DEBUG
[pantavisor] 135 DEBUG   -- [controller]: going to state = STATE_WAIT                                                                                                                                                                       
[pantavisor] 135 INFO    -- [debug]: shell started with pid 1072                                     
[pantavisor] 137 DEBUG   -- [controller]: going to state = STATE_DEBUG                            
[pantavisor] 137 DEBUG   -- [controller]: going to state = STATE_WAIT                                      
[pantavisor] 139 DEBUG   -- [controller]: going to state = STATE_DEBUG                                                                   
[pantavisor] 139 DEBUG   -- [controller]: going to state = STATE_WAIT
```

It also has the disadvantage that, like this, we have to process again the whole WAIT state and we might delay ctrl requests. Unless we are trying to avoid ctrl requests, sure.

Besides that, we are trying to get rid of most of the states in the future substituting periodic tasks such as this with events.

debug.c
```diff
@@ -46,44 +51,184 @@
...
+static struct timer shell_timer;
+static bool shell_session = false;
+static bool shell_timeout_active = false;
 
-void pv_debug_start_shell()
+static uint64_t pv_debug_timeout_elapsed_sec()
+{
+	struct timer_state state = timer_current_state(&shell_timer);
+	if (!state.fin) {
+		pv_log(DEBUG, "shell timeout to reboot: %d ", state.sec);
```
anibalportero, line 62: This is always printing 0:

```
[pantavisor] 1247 DEBUG  -- [controller]: going to state = STATE_WAIT
[pantavisor] 1249 DEBUG  -- [controller]: going to state = STATE_DEBUG
[pantavisor] 1249 DEBUG  -- [debug]: shell timeout to reboot: 0 
[pantavisor] 1249 DEBUG  -- [controller]: going to state = STATE_WAIT
[pantavisor] 1251 DEBUG  -- [controller]: going to state = STATE_DEBUG
[pantavisor] 1251 DEBUG  -- [debug]: shell timeout to reboot: 0 
[pantavisor] 1251 DEBUG  -- [controller]: going to state = STATE_WAIT
[pantavisor] 1253 DEBUG  -- [controller]: going to state = STATE_DEBUG
[pantavisor] 1253 DEBUG  -- [debug]: shell timeout to reboot: 0
```

ctrl.c
```diff
@@ -796,6 +796,14 @@ static int pv_ctrl_check_command(int req_fd, struct pv_cmd **cmd)
 		goto error;
 	}
 
+	if (!pv_config_get_bool(PV_DEBUG_SHELL) &&
```
anibalportero, line 799: I would expect another error message for when user is trying to defer but nothing can be done. That is, when timeout is not yet activated.

pantavisor.c
```diff
@@ -768,6 +775,20 @@ static pv_state_t _pv_command(struct pantavisor *pv)
 		}
 		pv->remote_mode = true;
 		break;
+	case CMD_DEFER_REBOOT:
+		if (!pv_config_get_bool(PV_DEBUG_SHELL)) {
```
anibalportero, line 779: do we want to do allow this when timeout is still not active?

pantavisor.c
```diff
@@ -768,6 +775,20 @@ static pv_state_t _pv_command(struct pantavisor *pv)
...
+			goto out;
+		}
+		if (strlen(cmd->payload) == 0)
+			goto out;
+
+		pv_log(DEBUG, "defer reboot command received, new timeout '%s'",
+		       cmd->payload);
+
+		pv_debug_defer_reboot_shell(cmd->payload);
+
```
anibalportero, line 791: you need a break here or it will print the "unknown command... " message down a couple lines

debug.c
```diff
@@ -46,44 +51,184 @@
...
+	read(con_fd, &c, sizeof(c));
+	close(con_fd);
 
-	if (c[0] == 'd' || pv_config_get_bool(PV_DEBUG_SHELL_AUTOLOGIN))
+	if (c[0] == '\n') {
 		shell_pid =
 			tsh_run("/sbin/getty -n -l /bin/sh 0 console", 0, NULL);
+
+		shell_session = true;
+		pv_wall("new shell session");
```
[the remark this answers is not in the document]
fortizc: yeah, and maybe could be useful to add the console pid?  @anibalportero

pantavisor.c
ferlzc: Oh yes, I forgot! thank for point that out!

pantavisor.c
ferlzc: I removed the PV_STATE_DEBUG and move the logic to PV_STATE_WAIT. I think this implementation is much better know.

debug.c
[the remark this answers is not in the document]
ferlzc: I rename this function to `pv_debug_shell().` This function take care of initialize the `/bin/sh` if the <ENTER> is pressed or PV_DEBUG_SHELL_AUTOLOGIN is enabled.
The fuction also check for potencial errors and treats then.  This function also track if the there is reboot timeout issue and tracks to see if the timeout is reached.
I moved the function from `bool` to `int`.

debug.c
```diff
@@ -46,44 +51,184 @@
...
+	if (shell_session && !once) {
+		debug_timeout = pv_config_get_int(PV_DEBUG_SHELL_TIMEOUT);
+		timer_start(&shell_timer, debug_timeout, 0,
+			    RELATIV_TIMER);
+
+		once = true;
+		shell_timeout_active = true;
+		pv_log(INFO, "shell timeout started with %d secs",
+		       debug_timeout);
+		pv_wall("System will reboot in 60 seconds, to defer reboot:"
```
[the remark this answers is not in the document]
ferlzc: done it!

wall.c
ferlzc: done!

debug.c
```diff
@@ -46,44 +51,184 @@
...
+
+static int is_shell_alive(void)
+{
+	if (shell_pid <= 0)
+		return 0;
+	// process is alive
+	if (kill(shell_pid, 0) == 0)
+		return 1;
+	// process is not alive or error occurred
+		return 0;
```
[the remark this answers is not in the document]
ferlzc: I guess i've fixed, please check!

debug.c
ferlzc: Fixed that! I was doing:

```
	struct timer_state state = timer_current_state(&shell_timer);
	if (!state.fin) {
		pv_log(DEBUG, "shell timeout to reboot: %d ", state.sec);
		return state.sec;
	}
	
```
which prints the wrong type. Funny enough on the Raspberry pi 3(arm64) it works but not on i.mx6ull which is arm32.

I used to print uint64_t which is the type of state.sec, like this:

`pv_log(DEBUG, "shell timeout to reboot: %"PRIu64, state.sec);`

debug.c
```diff
@@ -46,44 +51,184 @@
...
+		pv_log(INFO, "shell timeout started with %d secs",
+		       debug_timeout);
+		pv_wall("System will reboot in 60 seconds, to defer reboot:"
+			"run 'pventer -c <container-name> pvcontrol defer-reboot [new timeout]'");
+		pv_wall_ssh_users("System will reboot in 60 seconds, to defer reboot:"
+			"run 'pventer -c <container-name> pvcontrol defer-reboot [new timeout]'");
+
+		return true;
+	}
+	pv_log(DEBUG, "start timeout : shell session %d, once: %d , shell_timeout_active: %d",
```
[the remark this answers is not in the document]
ferlzc: I was using this pv_log to debug during development. I forgot to clean it afterwards.

debug.c
fortizc: > Funny enough on the Raspberry pi 3(arm64) it works but not on i.mx6ull which is arm32.

That's because the size of a integer is variable on different architectures. 

The safest way to fix this is to cast the number to a intmax_t or uintmax_t, as appropriate, and use the right placeholder %jd or %ju

pantavisor.c
ferlzc: "Maybe it's better to avoid changing the timeout if it is not active, because we have the `PV_DEBUG_SHELL_TIMEOUT` configuration, which can be set by the user. Changing the timeout value beforehand, if the timeout is already active, could lead to confusion. 
I'm processing this inside the `pv_debug_defer_reboot_shell` function.

debug.c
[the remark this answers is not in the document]
anibalportero: It's good to have it I guess

debug.c
ferlzc: thanks!  uintmax_t fixed!

PR #511, opened 2025-05-27
WIP: New binary protocol to logserver with a small header

fortizc:
New binary protocol for logserver:
* Include a small header to ensure to identify the messages from other data
* Limit the size of the log messages (see LOGSERVER_MAX_MSG_LEN)
* Improve how messages are parsed
* All new messages (commands, fd subscriptions and logs) use the new protocol

### DIFF
--- a/logserver/logserver.c
+++ b/logserver/logserver.c
@@ -19,6 +19,9 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
  * SOFTWARE.
  */
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE
+#endif
 
 #include <linux/limits.h>
 #include <sys/epoll.h>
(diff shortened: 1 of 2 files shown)


fortizc:
> what is the idea of backward compatiblity (e.g. none?). would we still support clients that log output?

The current patch, still support the old protocol, but should be completely deprecated at some point.

To me, the ideal solution will be to implement a complete binary format, I mean the buffer field in the struct is a bunch of string separated by `\0`, but everything should be binary. Well if we are 100% that isn't necessary to support the current format, we can do it now.

fortizc:
Regarding the last comment, when I said "complete binary format" I mean something like 
```c
struct logserver_msg {
        int version;
        int len;
        int level;
        char platform[64];
        char src[64];
        char msg[512];
};
```

fortizc:
> After landing, it would be nice to document this somewhere. Both legacy (marked as deprecated) and current formats. Maybe here https://docs.pantahub.com/logserver-sockets/?

Yeah, I think isn't ready to merge, I'm thinking to convert this to something which support RFC 5424 which is the current syslog format, so we could have compatibility with all the standard functions to log like [syslog(3)](https://man7.org/linux/man-pages/man3/syslog.3.html) so eventually that could happen here or in a different PR I will let you know about the new plans as soon as I can move to this task again

PR #513, opened 2025-06-03
Fix pvtx command

fortizc:
Fix some rough edges in pvtx
* Remove some debug prints
* Fix bugs linking bsp under specific circumstances
* Fix deploy behaviour to match with the old pvtx script


asac:
commit message has proyect not project :)

PR #516, opened 2025-06-11
Remove the libthttp dependency from pvtx

fortizc:
This means that now PVTX has its own version of jsmn.h (https://github.com/zserge/jsmn) 
The current version is 25647e692c7906b96ffd2b05ca54c097948e879c


asac:
lgtm

PR #519, opened 2025-06-17
Add dynamic buffer to read and write operations

fortizc:
Add dynamic buffer to read and write operations

Dynamic buffers are implemented to send and receive objects and json from pv-ctrl (pvtx_ctrl) and to read and write objects to folders (pvtx_txn), those buffers can be set to use different sizes, although they have minimum and maximum values. The following table shows a small summary for each one:

| variable name        |        operation       |    minimum   |     maximum     |
|----------------------|------------------------|--------------|-----------------|
| PVTX_CTRL_BUF_SIZE   | for pv-ctrl operations | 16384B (16K) | 10485760 (10M)  |
| PVTX_OBJECT_BUF_SIZE | for write/read objects | 512B         | 10485760B (10M) |

### DIFF
--- a/pvtx/pvtx_buffer.c
+++ b/pvtx/pvtx_buffer.c
@@ -1,0 +1,90 @@
+/*
+ * Copyright (c) 2025 Pantacor Ltd.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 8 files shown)


pvtx/pvtx_buffer.c
```diff
@@ -0,0 +1,125 @@
...
+
+int pv_pvtx_buffer_mempcpy(struct pv_pvtx_buffer *buf, void *src, size_t size)
+{
+	if (size > buf->size)
+		return -1;
+
+	buf->cur = mempcpy(buf->data, src, size);
+	return 0;
+}
+
```
ferlzc, line 117: is there a missing return here ?

ferlzc: In the following functions:


int pv_pvtx_buffer_mempcpy(struct pv_pvtx_buffer *buf, void *src, size_t size)
int pv_pvtx_buffer_memcpy(struct pv_pvtx_buffer *buf, void *src, size_t size)
int pv_pvtx_buffer_mempcpy(struct pv_pvtx_buffer *buf, void *src, size_t size)
int pv_pvtx_buffer_memccpy(struct pv_pvtx_buffer *buf, void *src, int c,
			   size_t size)
			   
void *src is not changed inside the function scope, so I think we can use const void *src  insted void *src for safety concerns.

pvtx/pvtx_buffer.c
```diff
@@ -0,0 +1,125 @@
...
+int pv_pvtx_buffer_memccpy(struct pv_pvtx_buffer *buf, void *src, int c,
+			   size_t size)
+{
+	if (size > buf->size)
+		return -1;
+
+	buf->cur = memccpy(buf->data, src, c, size);
```
ferlzc, line 124: is there a missing return in

int pv_pvtx_buffer_memccpy(struct pv_pvtx_buffer *buf, void *src, int c,
			   size_t size)
{
	if (size > buf->size)
		return -1;

	buf->cur = memccpy(buf->data, src, c, size);
}
?

fortizc: I accidentally upload on this PR an older version of pvtx_buffer, all those function (mem*cpy) will be removed, thx to catch this

PR #522, opened 2025-07-14
fix: escape bad formatted JSON devmeta

anibalportero:
When storing bad formatted JSON in device metadata, Pantavisor was unable to upload it as the HTTP request was bad formatted too. Escaping bad formatted JSON in these cases will solve this.

It also affected pvcontrol, which is now returning a JSON that contains escaped bad formatted JSONs too.

### DIFF
--- a/metadata.c
+++ b/metadata.c
@@ -914,8 +914,7 @@ int pv_metadata_upload_devmeta(struct pantavisor *pv)
 		char *val = pv_json_format(info->value, strlen(info->value));
 
 		if (key && val) {
-			// if value is a regular string
-			if (info->value[0] != '{') {
+			if (!pv_json_is_valid(info->value)) {
 				int frag_len = strlen(key) + strlen(val) +
 					       // 2 pairs of quotes
 					       2 * 2 +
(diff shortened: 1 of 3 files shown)


metadata.c
```diff
@@ -914,8 +914,8 @@ int pv_metadata_upload_devmeta(struct pantavisor *pv)
 		char *val = pv_json_format(info->value, strlen(info->value));
 
 		if (key && val) {
-			// if value is a regular string
-			if (info->value[0] != '{') {
+			if (!pv_json_is_valid(info->value)) {
+				// if value is a regular string
```
fortizc, line 918: Imo this comment now is useless `pv_json_is_valid()` is very explicit

utils/json.c
```diff
@@ -217,6 +217,27 @@ char *pv_json_format(const char *buf, int len)
...
+{
+	bool ret = false;
+
+	int tokc;
+	jsmntok_t *tokv = NULL;
+
+	if (jsmnutil_parse_json(json, &tokv, &tokc) < 0)
+		goto out;
+
+	if (!jsmnutil_get_object_keys(json, tokv))
```
fortizc, line 230: This second check is needed? and is our parser running in strict format? see (https://github.com/zserge/jsmn/blob/master/jsmn.h#L145)

anibalportero: Thanks for the comment. I'll switch that to jsmnutil_object_key_count, which only checks if first token is JSMN_OBJECT. Apparently, jsmnutil_parse_json will take any string as good and just set it as JSMN_PRIMITIVE.

PR #530, opened 2025-08-06
feat: add support for allowing options to dm-verity script

ferlzc:
- updated the syntax for veritysetup command
- add pantavisor global config to enable --panic-on-corruption to veritysetup


scripts/volmount/verity/dm
```diff
@@ -56,13 +61,23 @@ do_mount_squash() {
...
+		$ddev \
+		$dmname \
+		$hdev \
+		$rhash \
+		--hash=$hashalgo \
+		--data-block-size=$datablocksize \
+		--hash-block-size=$hashblocksize \
+		--salt=$shash \
+		$verity_opts; then
+	echo "ERROR: could not setup dm volume with veritysetup"
```
anibalportero, line 78: indent

scripts/volmount/verity/dm
```diff
@@ -47,6 +51,7 @@ do_mount_squash() {
 	dmmanifest=$traildir/_dm/$squashname.json
 	dmname=pv--dmverity--$container-${squashname%.*}
 	volpath=/volumes/$container/$squashname
+	verify_opts=$3
```
anibalportero, line 54: typo

scripts/volmount/verity/dm
```diff
@@ -27,6 +27,12 @@ volumesrc=$2
 container=$3
 volume=$4
 
+veritysetup_options=$PV_VOLMOUNT_DM_EXTRAARGS
+# fixed variables for veritysetup
+hashalgo=sha256
+hashblocksize=4096
+datablocksize=4096
```
ferlzc, line 34: I add those values that veriftsetup needs as a constant as example. We can also setup an config to change that via pantavisor config.

init.c
```diff
@@ -426,6 +426,10 @@ int main(int argc, char *argv[])
 	// this might override the configuration
 	parse_commands(argc, argv);
 
+	if (pv_config_get_str(PV_VOLMOUNT_DM_EXTRAARGS) != NULL)
```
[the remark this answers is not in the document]
ferlzc: Ok I move as argument on volume.c. Make sense to be used there.

config.c
```diff
@@ -255,6 +255,8 @@ static struct pv_config_entry entries[] = {
 	  .value.i = 120 },
 	{ BOOL, "PV_UPDATER_USE_TMP_OBJECTS", PV | OEM | RUN, 0,
 	  .value.b = false },
+	{ STR, "PV_VOLMOUNT_DM_EXTRAARGS", LEVEL_SYSCTL, 0, false,
```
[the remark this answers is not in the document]
ferlzc: I used LEVEL_SYSCTL because of this:

`#define LEVEL_SYSCTL PV | OEM`

so in the end is `PV | OEM `

There is another reason for `LEVEL_SYSCTL` insted of `PV | OEM `?

Anyway I'm moving to `PV | OEM`

`PV_VOLMOUNT_DM_EXTRA_ARGS` -> much better!

scripts/volmount/verity/dm
ferlzc: @asac  should we add configs for these variables ?

scripts/volmount/verity/dm
```diff
@@ -27,6 +27,12 @@ volumesrc=$2
 container=$3
 volume=$4
 
+veritysetup_options=$PV_VOLMOUNT_DM_EXTRAARGS
```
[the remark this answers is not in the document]
ferlzc: Yeah veritysetup is will fail when it gets a invalid option. Maybe check all the options before calling veritysetup is too much because we are not using the error code.

volumes.c
```diff
@@ -232,20 +233,42 @@ int pv_volume_mount(struct pv_volume *v)
...
 					     strlen("%s umount %s %s %s") + 1));
-			sprintf(command, "%s mount %s %s %s", script, path,
-				partname, name);
+
+			if (verity_options == NULL)
+				sprintf(command, "%s mount %s %s %s", script,
+					path, partname, name);
+			else
+				sprintf(command, "%s mount %s %s %s %s", script,
+					path, partname, name, verity_options);
```
fortizc, line 265: For all this piece of code, I really recommend to use `asprintf()`

ferlzc: Cool! `asprintf()` fits great for this case! This part of the code was the way it was done before, I've just add some new conditions. I'm refactoring to `asprintf()`

PR #532, opened 2025-08-12
fix: add debug shell in early boot

ferlzc:


tools/pvlog
```diff
@@ -0,0 +1,42 @@
...
+			usage
+			exit 1
+			;;
+	esac
+done
+
+if [ -n "$rev" ]; then
+	log="/pv/logs/$rev/pantavisor/pantavisor.log"
+else
+	log="/pv/logs/0/pantavisor/pantavisor.log"
```
fortizc, line 39: This is wrong, if your device is not connected to the cloud from the beginning, there is no rev 0. The default path should be `/pv/logs/current/pantavisor/pantavisor.log` now the problem is that the link (current) could not exist for 2 reasons:

1. You are at a very early stage of the start and the system does not yet create the link
2. Something goes wrong and the system cannot create the link

ferlzc: I used 0 because the `current`  wasn't there in the begining. I will double check it but I think that was the case.

tools/pvlog
```diff
@@ -0,0 +1,42 @@
+#!/bin/sh
+
+rev=""
+
+usage() {
+	echo "USAGE: pventer [--rev|-r <REV>] [--help|-h]"
```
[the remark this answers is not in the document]
ferlzc: thanks for spoting that!

anibalportero (commented):
I tested the change. Both early and regular shell are running fine.

PR #539, opened 2025-09-03
Add detection of objects in queue folders

fortizc:
With this PR the all the packages could be decompressed in a directory named `NNN__ abc` where `NNN` are numbers and `abc` is a text to identify the folder. With this configuration, a queue can be created with `pvtx queue new` and then, when `pvtx queue process` be called, pvtx will process all the json files and move the objects to the object folder.

The order how pvtx will process the folders depends on the `NNN` prefix, pvtx always process the lower numbers first. It's very important to notice that the folders needs the three digits, and for small numbers, they should be padded with 0, ej: 000__pgk1 001_pkg2, and so on


ferlzc:
LGTM. The only comment I may add is to document somewhere the folders name with three digits and the padding for small numbers.

PR #543, opened 2025-09-09
fix: add verification on asprintf call on volume mount

ferlzc:


volumes.c
```diff
@@ -240,15 +240,21 @@ int pv_volume_mount(struct pv_volume *v)
...
 
-			asprintf(&umount_cmd, "%s umount %s %s %s", script,
-				 partname, path, name);
+			ret = asprintf(&umount_cmd, "%s umount %s %s %s",
+				       script, partname, path, name);
+
+			if (ret < 0) {
+				pv_log(ERROR,
+				       " cannot allocate memory for mount/umount command");
+				return ret;
```
fortizc, line 256: At this point `handler` could have been allocated, so maybe it's better to "goto out" instead exit at this point using `return`

PR #659, opened 2026-03-19
feat: add hooks for updates

fortizc:
This feature adds the possibility to add scripts at `$PV_INSTALL_FULL_PVLIBDIR/hooks/system.d` (commonly `/usr/lib/pantavisor/pv/hooks/system.d`) to run at some system events:
* system-start
* system-before-install-update
* system-after-install-update
* system-install-update
* system-boot-done
* system-done

On each call, the scripts are called with the following environment variables set:
* PV_TRY: current pv_tryt 
* PV_REV: current rev
* PV_TRYBOOT: true/false
* PV_OBJ_STORAGE: path to object storage 
* PV_TRAILS_STORAGE: path to trails
* PV_STATUS: current revision status.
* PV_OP: one of the mentioned events (step-*)

Also, a couple of mock script are included: `00-hook-mock` and `99-hook-mock` those files look for a config file in `$PV_INSTALL_FULL_PVLIBDIR/hooks-mock-conf/00-hook-mock/REV.cfg` and `$PV_INSTALL_FULL_PVLIBDIR/hooks-mock-conf/00-hook-mock/REV.cfg` respectively. The name `REV.cfg` means that each hook look a config named as the current revision, so if the current revision is `locals/succeed-system-boot-done` the hooks will use `$PV_INSTALL_FULL_PVLIBDIR/hooks-mock-conf/00-hook-mock/locals/succeed-system-boot-done.cfg`, and behaves following the configuration, which is pretty simple: pairs of `PV_OP` and an expected behaviour where the options are:
* OK: the hook finish with exit status == 0
* ERROR: the hook finish with exit status == 1
* REBOOT: the hook calls systems reboot (pvcontrol cmd reboot)


docs/overview/hooks.md
```diff
@@ -0,0 +1,89 @@
...
+- Only **regular files** with the **executable bit** (`S_IXUSR`) set are run; directories, symlinks, and non-executable files are silently skipped.
+- Naming convention: use a numeric prefix (e.g. `10-my-hook`, `50-notify`) to control execution order.
+
+## Hook Points
+
+Each hook point corresponds to a specific moment in Pantavisor's lifecycle. The active hook point is communicated to hook scripts via the `PV_OP` environment variable.
+
+| `PV_OP` value | When it fires |
+|---|---|
+| `system-start` | Early during Pantavisor initialization, before containers are started. |
```
anibalportero: link to state machine in pantavisor-architecture.md (see around about how other links are done with the ../../../ etc.)

docs/overview/hooks.md
```diff
@@ -0,0 +1,89 @@
...
+| `system-before-install-update` | Before the bootloader writes the new revision to its environment. Only fired on bootloaders that support separate before/after install hooks. |
+| `system-after-install-update` | After the bootloader has written the new revision. Only fired on bootloaders that support separate before/after install hooks. |
```
anibalportero: link to update.md

hook/hooks.h
```diff
@@ -0,0 +1,38 @@
...
+
+int pv_hooks_set_env(const char **env, int size);
+int pv_hooks_set_var(const char *key, const char *value);
+void pv_hooks_unset_var(const char *key);
+void pv_hooks_unset_env(const char **env, int size);
+void pv_hooks_unset_default_env(const char **extra_env, int size);
+void pv_hooks_set_default_env(const char *pv_op, const char *pv_rev,
+			      const char *pv_try, const char **extra_env,
+			      int size);
+int pv_hooks_run(const char *dir, bool log);
```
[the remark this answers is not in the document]
fortizc: Same as previous comment, the idea is to get something more generic, it seems we could implement more hooks in the future

paths.h
```diff
@@ -62,6 +62,9 @@ void pv_paths_pv_log_plat(char *buf, size_t size, const char *rev,
 void pv_paths_pv_log_file(char *buf, size_t size, const char *rev,
 			  const char *plat, const char *name);
 
+#define UPDATE_HOOKS "hooks"
+void pv_paths_pv_update_hooks(char *buf, size_t size);
```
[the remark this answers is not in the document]
anibalportero: ok

utils/tsh.c
```diff
@@ -104,16 +105,43 @@ static pid_t _tsh_exec(char **argv, int wait, int *status, int stdin_p[],
...
+
+		} else if (ret == 0) { // wait only if we blocked SIGCHLD
+
+			timeout = timeout <= 0 ? TSH_DEFAULT_TIMEOUT : timeout;
+
+			struct timespec tm = {
+				.tv_sec = timeout,
+				.tv_nsec = 0,
+			};
+
```
asac: the new line strucutre looks odd...

paths.h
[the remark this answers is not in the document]
anibalportero: I still don't get the update part if you are returning /usr/lib/pantavisor/pv/hooks/ and you are storing hooks that have nothing to do with updates. Also system.d would be a nice #define here instead of UPDATE_HOOKS, which is not used outside paths.c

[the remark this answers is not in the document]
fortizc: yes, I guess you're right, the "update" comes from the original idea, now things has changed a bit, I guess remove that "update" is a good idea

bootloader.c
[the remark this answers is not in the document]
fortizc: Mmm... I see you're talking about `_finish_update_installation()` where this function is called. It should be interesting to get something like a `int pv_bootloader_install_update(char *rev, char *error_msg)` where the underlying error will be specified. This is valid for many parts in Pantavisor, maybe we should think in something more "standard" to solve this case

docs/overview/hooks.md
```diff
@@ -0,0 +1,91 @@
...
+## Environment Variables
+
+Pantavisor sets the following environment variables before executing every hook:
+
+| Variable | Description |
+|---|---|
+| `PV_OP` | Name of the [hook point](#hook-points) currently running (see table above). |
+| `PV_REV` | ID of the currently running [revision](revisions.md). |
+| `PV_TRY` | ID of the revision being attempted for the next boot. Empty string if not in a try-boot or once the try-boot has been committed. For update hooks (`system-*-install-update`), this is the incoming revision being installed. |
+| `PV_TRYBOOT` | `"true"` if the device booted into a trial revision that has not yet been committed, `"false"` otherwise. Always `"true"` for `system-boot-done`. |
```
czesiekbis: Why have it always "true" for system-boot-done?  Isn't the fact that system-boot-done is called enough? 
This special casing introduces  conflicting wording in descriptions of system-boot-done and PV_TRYBOOT

fortizc: yeah, that's true, that last part isn't necessary, thanks

CMakeLists.txt
```diff
@@ -27,6 +27,7 @@ set(PANTAVISOR_PVTX_STATIC OFF CACHE BOOL "build pantavisor pvtx static")
 set(PANTAVISOR_CLANG_FORMAT_CHECK OFF CACHE BOOL "run clang-format check on build")
 set(PANTAVISOR_XCONNECT OFF CACHE BOOL "Build pantavisor xconnect")
 set(PANTAVISOR_DCP_BLOB_CREATE OFF CACHE BOOL "Build dcp-blob-create tool for DCP trusted key migration")
+set(PANTAVISOR_MOCK_HOOKS ON CACHE BOOL "Add mock test for Pantavisor hooks")
```
asac: **Must-fix**: default this to `OFF`. With `ON` as the default, every shipped image installs `00-hook-mock` / `99-hook-mock` plus the 24 `*.cfg` files. The mock calls `pvcontrol cmd reboot` whenever the running revision name matches `reboot-system-*`, so a test/QA revision name on a fielded device triggers an unintended reboot loop. Mock should be opt-in for CI/dev configs only.

hook/hooks.h
```diff
@@ -0,0 +1,38 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#ifndef PV_HOOK_HOOKS_H
+#define PV_HOOK_HOOKS_H
+
```
asac, line 25: API is awkward: `(const char **env, int size)` where `env` is a flat KV array but `size` is the *pair* count. Call sites already look odd:

```c
const char *env[] = { "PV_TRYBOOT", "true" };
pv_hooks_set_default_env("system-boot-done", rev, "", env, 1);
```

Prefer `(const char *kv[][2], int n)` or a `struct { const char *key, *val; }` array. Also `pv_hooks_set_default_env` returns `void`, so `setenv()` failures (ENOMEM) silently disappear — consider returning an int.

hook/hooks.c
```diff
@@ -0,0 +1,293 @@
...
+void pv_hooks_unset_var(const char *key)
+{
+	if (!key || !secure_getenv(key))
+		return;
+
+	unsetenv(key);
+	pv_log(DEBUG, "removing environment var %s", key);
+}
+
+void pv_hooks_unset_env(const char **env, int size)
```
asac, line 88: `secure_getenv` is the wrong primitive here. It returns NULL under AT_SECURE; Pantavisor isn't setuid so today it's equivalent to `getenv`, but it's confusing and brittle. Use `getenv`.

hook/hooks.c
```diff
@@ -0,0 +1,293 @@
...
+	char dir[PATH_MAX] = { 0 };
+	SNPRINTF_WTRUNC(dir, sizeof(dir), "%s/%s", base, dirname);
+
+	struct dirent **entry = NULL;
+	int count = scandir(dir, &entry, NULL, alphasort);
+
+	if (count < 0) {
+		pv_log(DEBUG, "couldn't open directory %s: %s", dir,
+		       strerror(errno));
+		return -1;
```
asac: Treat `ENOENT` (directory absent) as success. On upgrades from a pre-hooks install the dir may not exist, and right now that returns -1 and blocks every update via `system-before-install-update`. The `install(DIRECTORY ...)` in CMake only protects fresh installs.

hook/hooks.c
```diff
@@ -0,0 +1,293 @@
...
+				pv_log(ERROR, "hook %s failed status: %d", path,
+				       WEXITSTATUS(wstatus));
+				ret = -1;
+				goto out;
+			} else if (WIFEXITED(wstatus)) {
+				pv_log(DEBUG, "hook %s succeeded", path);
+				ret = 0;
+			} else if (WIFSIGNALED(wstatus)) {
+				pv_log(ERROR, "hook execution signaled %s: %d",
+				       path, WTERMSIG(wstatus));
```
asac, line 254: **No timeout.** `tsh_run_io` / `tsh_run_logserver` block until the script exits. A buggy hook (`while :; do :; done`) hangs PID1. For `system-start` the tryboot path eventually rescues us *only* on bootloader backends that decrement the tryboot counter independently — on backends where Pantavisor decrements it, a hung hook defeats rollback. Update hooks block update progress regardless. Need a per-hook timeout with SIGKILL on expiry.

hook/hooks.c
```diff
@@ -0,0 +1,293 @@
...
+
+	return status;
+}
+
+void pv_hooks_set_default_env(const char *pv_op, const char *pv_rev,
+			      const char *pv_try, const char **extra_env,
+			      int size)
+{
+	if (!pv_rev)
+		pv_rev = pv_bootloader_get_rev();
```
asac, line 128: `get_current_status()` always reads `current`, even for update hooks where the operator likely wants the *incoming* status. Either document this clearly in `hooks.md` (it isn't currently) or set it correctly per hook point.

bootloader.c
```diff
@@ -144,12 +145,53 @@ int pv_bootloader_install_update(char *rev)
 	       rev);
 
 	if (ops->install_update) {
+		char *stored_rev = ops->get_env_key("pv_rev");
```
asac: `ops->get_env_key` is not null-checked here (also at the post-install call and at line ~234 in `pv_bootloader_post_commit_update`). If a backend lacks `get_env_key`, the install path crashes. Match the existing `if (ops->install_update)` guard pattern.

bootloader.c
```diff
@@ -190,6 +232,16 @@ int pv_bootloader_post_commit_update(const char *rev)
 
 	pv_log(INFO, "revision %s commited", rev);
 
+	const char *env[] = { "PV_TRYBOOT", "true" };
```
asac: `PV_TRYBOOT` redundancy: caller passes `extra_env = {"PV_TRYBOOT","true"}`, which overwrites the value `pv_hooks_set_default_env` already derived from `pv_bootloader_trying_update()`. Either rely on the helper or skip it — not both.

Related concern at `system-after-install-update` (around line 167): `pv_bootloader_set_try()` has not yet run at hook time, so the helper-derived `PV_TRYBOOT` may be `false` even though we're squarely in a tryboot install. Pin down the contract and make the code match the docs.

bootloader.c
```diff
@@ -338,4 +393,4 @@ static int pv_bl_early_init(struct pv_init *this)
 struct pv_init pv_init_bl = {
 	.init_fn = pv_bl_early_init,
 	.flags = 0,
-};
+};
```
asac: Accidental "no newline at end of file" — please restore. Same in `pantavisor.c` and `state.c`.

asac (commented):
Concept and docs are solid. A few items to address before merge — leaving out sandboxing concerns since hooks ship in the initramfs and are secureboot-covered (no third-party drop-in path).

## Must-fix

**1. `PANTAVISOR_MOCK_HOOKS` defaults to `ON`** — `CMakeLists.txt:30`

This installs `00-hook-mock` / `99-hook-mock` and the 24 `*.cfg` files into every shipped image. The mock executes `pvcontrol cmd reboot` whenever the running revision name matches `reboot-system-*`. A test/QA revision name on a fielded device would trigger an unintended reboot loop. Default must be `OFF`; only enable in CI/dev configs.

**2. No timeout on hook execution** — `hook/hooks.c:243-272`

`tsh_run_io` / `tsh_run_logserver` block until the script exits. A buggy hook (`while :; do :; done`) hangs PID1. For `system-start` the tryboot/rollback path saves us *if* the bootloader-side counter eventually fires the watchdog — but on backends where Pantavisor itself decrements the tryboot counter, a hung hook defeats rollback. Update hooks block update progress regardless. Add a per-hook timeout (configurable, with a hard ceiling) and SIGKILL on expiry.

**3. Missing-directory case fails every hook invocation** — `hook/hooks.c:188-192`

If `system.d/` doesn't exist (e.g., upgrade from a pre-hooks install), `scandir` returns -1 and `pv_hooks_run` returns -1, blocking updates. Treat `ENOENT` as "no hooks → success." The `install(DIRECTORY ...)` only protects fresh installs.

## Documentation gap (important)

The hooks-are-part-of-the-revision contract is the keystone of the design but isn't spelled out in `docs/overview/hooks.md`. Please add:

- A short **"Hooks are part of the revision"** section: the hook ships with the rev, runs at try-boot, and a failure rolls back via the normal tryboot path.
- For `system-start`: failure is fatal → reboot → tryboot counter → rollback to last-good. Treat it as a **gate** on the revision, not a side-effect.
- The asymmetry with `system-done` (logged-only) is intentional and worth calling out explicitly.
- The `system-before/after-install-update` row says "Only fired on bootloaders that support separate before/after install hooks." List which backends fall on each side, or link to the matrix.
- "Install the script ... is done in build time in the CMakeLists.txt file" — operators usually ship hooks via BSP overlays, not by editing pantavisor's CMakeLists. Reframe around the BSP/integration side.

Also update `docs/overview/pantavisor-architecture.md` state-machine list to mention the other hook points (currently only `system-start` is listed there).

## Security

**4. `secure_getenv` is the wrong primitive** — `hook/hooks.c:88` (`pv_hooks_unset_var`)

`secure_getenv` returns NULL under AT_SECURE. Pantavisor isn't setuid so today it's equivalent to `getenv`, but it's confusing and brittle. Use `getenv`.

**5. Path-traversal vector in mock hook** — `scripts/system-hooks/hook-mock:3`

`conf=".../$(basename "$0")/$PV_REV.cfg"`. If `PV_REV` ever contains `..` or `/`, the script reads arbitrary files. Mock-only, low impact, but a one-line `case "$PV_REV" in *..*|*/*) exit 0;; esac` guard is cheap.

## Coding issues

Filed as inline comments on the relevant lines.

## Nits

- `hook-mock`: leading `printf` lacks `\n`; the conditional path then `echo`s. Minor cosmetic.
- `state.c` `system-done` fires on transition into `PLAT_READY` — correct given the `s->status == status` early-return, but worth a one-line comment.
- Commit hygiene: the lead commit body is a stack of 11 mini messages from a rebase squash (`fix:`, `feat:`, `fix:`, ...). Please squash into a clean single commit (or a small ordered series), per Conventional Commits.

bootloader.c
```diff
@@ -35,6 +35,7 @@
 #include "config.h"
 #include "init.h"
 #include "pantavisor.h"
+#include "hook/hooks.h"
```
asac, line 38: `#include "hook/hooks.h"` is the only `subdir/` include in this file. Other modules use flat includes (mirroring how the existing source tree is organized). Pick a convention — either move the header up or update other includes.

hook/hooks.c
fortizc: This is wrong, by default any call to tsh functions has 60 sec of timeout

hook/hooks.c
fortizc: so `get_current_status()` should have the following logic, I guess:
```
if (pv_try)
    pv_storage_get_rev_progress(pv_try)
else
    pv_storage_get_rev_progress(pv_rev) // "current"
```
Please confirm if this is better (I guess so)

bootloader.c
fortizc: I don't think so, the line just below include "utils/str.h"

hook/hooks.c
asac: lets first understand what value makes sense for each hook here and then see what the right answer is.

CMakeLists.txt
fortizc: done 9b2c7e70a371f76fdea2fa5caa45e9ec27118de0

hook/hooks.h
fortizc: done 6cf0b37571ea4ef92f76253674871b3f068b78af

hook/hooks.c
fortizc: done 909b6d4d3c411e14071e486d3d3ffe4f193e4ba1

hook/hooks.c
fortizc: done 79a2382b823fa384efd142c89a24d4eb727ddfc8

bootloader.c
fortizc: done 2b8b9c6d641ae2467d0ee7fc3ed745eded2ae848

bootloader.c
fortizc: done 2b8b9c6d641ae2467d0ee7fc3ed745eded2ae848

hook/hooks.c
fortizc: done dd9b3f4865f58c79460feff08a0c39e79296cdea

bootloader.c
fortizc: done 4d579a5b3c70178f8707934a3737af3f9efa90ad

PR #678, opened 2026-04-07
fix(pvtx): fix tar parser bugs, memory leaks, and fd leak

fortizc:
# Fix tar parser correctness issues, memory/fd leaks, and improve I/O performance

### Bug fixes (`pvtx_tar.c`)

- **Long-name detection** (`get_name`): The previous implementation detected GNU tar long-name entries by comparing the entry name against the magic string `"/./@LongLink"`. The correct approach is to check `typeflag == 'L'`, which is what the GNU tar spec requires. The old code also contained a latent type bug: `char *buf[PVTX_TAR_BLOCK_SIZE]` (array of char pointers) instead of `char buf[PVTX_TAR_BLOCK_SIZE]` (byte array).

- **fd leak** (`pv_pvtx_tar_from_path`): If `pv_pvtx_tar_from_fd()` failed, the file descriptor opened by `from_path` was never closed.

- **Missing null-check** (`pv_pvtx_tar_from_fd`): `imp->from_fd()` could return NULL (e.g., on `gzdopen` failure) with no error propagation, leading to a NULL dereference later.

- **Zero-size entry capacity** (`pv_pvtx_tar_next`): The block-alignment rounding `(size | (BLOCK_SIZE - 1)) + 1` produced `BLOCK_SIZE` instead of `0` for zero-size entries, causing an unnecessary read of a padding block.

- **Free path** (`pv_pvtx_tar_free`): Replaced a `goto`-based null guard with nested `if` checks, making the ownership and cleanup logic explicit.

### Memory/allocation improvements

- `read_remaining_bytes()` previously allocated a heap buffer (size configurable via `PVTX_OBJECT_BUF_SIZE`, up to 10 MiB) through `pv_pvtx_buffer_from_env()`. Since the function only drains remaining bytes and discards them, a 4 KiB stack buffer is sufficient and removes an allocation that could fail.

- `pv_pvtx_tar_next()` now reads the tar header directly into the `pv_pvtx_tar_metadata` struct, eliminating an intermediate `char buf[]` + `memcpy`.

- `pv_pvtx_tar_content_read_object()` now issues a single `read_nointr()` for all remaining bytes instead of looping one block at a time, reducing syscall overhead for large objects.

### I/O performance (`pvtx_tar_formats.c`)

- Raw format: `posix_fadvise(POSIX_FADV_SEQUENTIAL)` is now called after `lseek`, advising the kernel to prefetch ahead aggressively.
- Gzip format: `gzbuffer(gz, 32768)` increases the zlib internal read buffer from the default 8 KiB to 32 KiB, reducing the number of underlying `read()` calls during decompression.

### DIFF
--- a/pvtx/pvtx_tar.c
+++ b/pvtx/pvtx_tar.c
@@ -25,7 +25,6 @@
 #endif
 
 #include "pvtx_tar_impl.h"
-#include "pvtx_buffer.h"
 #include "pvtx_tar.h"
 
 #include <unistd.h>
@@ -61,6 +60,8 @@ struct pv_pvtx_tar_metadata {
 	char prefix[155];
 };
 
+#define PVTX_TAR_TYPEFLAG_LONG_NAME 'L'
+
 extern struct pv_pvtx_tar_imp pv_pvtx_tar_raw;
 extern struct pv_pvtx_tar_imp pv_pvtx_tar_gzip;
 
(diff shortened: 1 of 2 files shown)


pvtx/pvtx_tar.c
```diff
@@ -182,13 +172,12 @@ ssize_t pv_pvtx_tar_content_read_object(struct pv_pvtx_tar_content *con,
...
-		ssize_t size = read_nointr(con->priv, p, PVTX_TAR_BLOCK_SIZE);
-		if (size > 0)
-			con->read += size;
-		else
-			break;
+	ssize_t remaining = con->cap - con->read;
+	if (remaining > 0) {
+		ssize_t n = read_nointr(con->priv, (char *)buf + con->read,
+					remaining);
+		if (n > 0)
```
asac, line 179: This replaces the old block-by-block loop with a single `read_nointr()` for all remaining bytes. While `read_nointr` retries on `EINTR`, a single `read()` is **not guaranteed to return all requested bytes** — short reads are valid (especially from gzip streams or pipes).

If `n < remaining`, the object gets silently truncated. Should this loop until `remaining` is fully consumed, or is there a guarantee from the underlying format implementations that short reads (other than EINTR) can't happen here?

pvtx/pvtx_tar.c
```diff
@@ -92,33 +93,22 @@ static ssize_t read_nointr(struct pv_pvtx_tar_priv *priv, void *buf,
...
 {
-	const char *long_link = "/./@LongLink";
-	if (strncmp(meta->name, long_link, strlen(long_link))) {
-		memccpy(name, meta->name, '\0', 100);
+	if (meta->typeflag == PVTX_TAR_TYPEFLAG_LONG_NAME) {
+		char buf[PVTX_TAR_BLOCK_SIZE] = { 0 };
+		read_nointr(tar->priv, buf, PVTX_TAR_BLOCK_SIZE);
+		memccpy(name, buf, '\0', NAME_MAX);
 		return;
 	}
```
asac, line 101: This reads exactly one 512-byte block for long names. GNU tar long-name entries can technically span multiple blocks if the name exceeds 512 bytes (the entry's `size` field tells how many bytes to read). `NAME_MAX` is typically 255 so it's fine in practice, but worth noting this is a pre-existing limitation — not introduced by this PR.

pvtx/pvtx_tar.c
```diff
@@ -333,6 +325,8 @@ struct pv_pvtx_tar *pv_pvtx_tar_from_path(const char *path,
...
```
asac, line 325: Nit: file is missing a trailing newline.

asac (commented):
Nice cleanup — the bug fixes are real and well-motivated. The long-name typeflag check, fd leak, NULL-check, and zero-size cap fix are all correct.

A couple of inline comments below. The main one is about `pv_pvtx_tar_content_read_object` — the single-read replacement for the old loop may silently truncate objects on short reads from gzip streams.

pvtx/pvtx_tar.c
asac: this seems to be ok actually.

pvtx/pvtx_tar.c
fortizc: As you mention in practice if fine. To implement something more complex we require read multiples blocks of the tar until find a real block with metadata, also we should allocate memory dynamically, so probably the cost of that is too high to not get a real improvement because the name cannot exceed the NAME_MAX size

PR #681, opened 2026-04-09
fix(platforms): unmount volumes when container transitions to STOPPED

asac:
## Summary

- Unmount platform volumes when a container transitions to PLAT_STOPPED
- Fixes mount accumulation (overmounting) during auto-recovery crash/restart cycles
- Each recovery cycle now gets clean mounts instead of stacking on top of existing ones

The fix is in `pv_platform_set_status()` which is the single chokepoint for all STOPPED transitions — crashes, force_stop, and failed starts.

## Test plan

- [x] Build with auto-recovery containers (e.g., `pv-example-recovery`)
- [x] Trigger multiple crash/recovery cycles
- [x] Verify `cat /proc/mounts | grep <container>` shows constant mount count (no accumulation)
- [ ] Verify normal shutdown/reboot still unmounts correctly

### DIFF
--- a/platforms.c
+++ b/platforms.c
@@ -41,6 +41,7 @@
 #include <sched.h>
 
 #include "platforms.h"
+#include "volumes.h"
 #include "paths.h"
 #include "wdt.h"
 #include "drivers.h"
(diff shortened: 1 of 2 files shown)


asac:
tested things in context of PR#682 that got rebased on this ...

PR #683, opened 2026-04-10
feat(event-log): unified event journaling subsystem

asac:
## Summary

Adds a structured event journaling system that records timestamped lifecycle events from all pantavisor subsystems into the revision's progress JSON payload. This gives admin tools and dashboards a machine-readable "black box" for understanding device behavior — which containers crashed, when recovery was attempted, whether cloud connectivity dropped, and how updates progressed.

- **Core module** (`event_log.h/c`): 128-entry ring buffer (~29 KB), 4 event types (`system`, `platform`, `update`, `pantahub`), config-gated via `PV_LOG_EVENTS`
- **Progress integration**: events array serialized into existing progress JSON, flows to disk (`.pv/progress`) and Hub (`PUT /steps/{rev}/progress`) via existing `report_cb` path
- **Hook points**: platform lifecycle (started/crashed/recovering/stable/stopped), update lifecycle (queued/downloading/boot/testing/committed/rollback), system state machine transitions, pantahub connectivity (online/offline/auth)
- **Config**: `log.events` / `PV_LOG_EVENTS` (bool, default true, `PV|OEM|RUN` levels) — can be disabled on flash-constrained devices or toggled at runtime
- **Documentation**: full overview at `docs/overview/event-journal.md`, config reference updated

## Dependencies

> This PR depends on pantavisor/pantavisor#677 (container lifecycle control API) being merged first. It includes those commits in the branch history.

## Hub-side change needed

The pantahub-base progress endpoint uses a typed Go struct (`StepProgress`). An `Events []StepEvent` field needs to be added to `trails/trailmodels/trailmodels.go` for events to be persisted in the Hub. Without this change, events are silently dropped by Go's JSON unmarshaler — but they still work locally (`.pv/progress` file + ctrl API). The changes are backwards-compatible in both directions; Hub should be deployed first.

## Design document

This was implemented against the unified pitch agreed by EngineerGem and EngineerCla:

<details>
<summary>FINAL-UNIFIED-PITCH-eventsubsystem.md</summary>

### Event Structure

```c
typedef enum {
    PV_EVENT_TYPE_SYSTEM,     // pantavisor state machine, daemons, storage, logserver
    PV_EVENT_TYPE_PLATFORM,   // container lifecycle
    PV_EVENT_TYPE_UPDATE,     // revision lifecycle (queued -> testing -> committed)
    PV_EVENT_TYPE_PANTAHUB    // cloud connectivity, auth, sync
} pv_event_type_t;

struct pv_event {
    time_t ts;                // Wall-clock timestamp
    pv_event_type_t type;
    char src[64];             // container name, rev number, subsystem, "pantahub"
    char event[32];           // started, crashed, recovering, stable, etc.
    char detail[128];         // space-separated key=value pairs
};
```

### Ring Buffer
- 128 entries, fixed-size global singleton
- Synthetic `wrapped` event on first overflow
- No deduplication (consumers summarize)

### Event Vocabulary

**platform** (src = container name): `started`, `ready`, `stable`, `crashed` (pid=N), `recovering` (attempt=N/M delay=Ns), `max_retries` (backoff=reboot|never|Ns), `stopped`, `user_stopped`, `user_started`, `user_restarted`

**update** (src = revision number): `queued` (retry=N/M), `downloading`, `reboot`, `transition`, `boot`, `testing` (mode=reboot|nonreboot), `committed`, `rollback`

**system** (src = subsystem): `state_change` (STATE_X -> STATE_Y)

**pantahub** (src = "pantahub"): `online`, `offline` (failed_reqs=N), `auth`, `auth_failed`

</details>

## Test plan

Testplan at `meta-pantavisor/TESTPLAN-event-subsystem.md` (21 tests). Basic appengine verification passed:

```
Test 1: Events array in progress file ............. PASS
Test 2: Events via ctrl API ........................ PASS
Test 3: Event types present ........................ PASS
Test 4: Event structure validation ................. PASS
Test 5: Progress consistency & chronological order . PASS
Test 6: state_change detail format ................. PASS
```

## Files changed

| File | Change |
|------|--------|
| `event_log.h` (new) | Event struct, ring buffer, API |
| `event_log.c` (new) | Ring buffer, JSON serialization, config gate |
| `config.h` + `config.c` | `PV_LOG_EVENTS` config entry |
| `CMakeLists.txt` | Add event_log sources |
| `update/update_progress.c` | Serialize events in `_ser_progress()` |
| `update/update.c` | Update lifecycle event hooks |
| `platforms.c` | Platform lifecycle event hooks |
| `state.c` | Auto-recovery event hooks |
| `pantavisor.c` | System state machine + boot event hooks |
| `ctrl/ctrl_containers_ep.c` | User action event hooks |
| `pantahub/pantahub_proto.c` | Hub connectivity event hooks |
| `docs/overview/event-journal.md` (new) | Full event journal documentation |
| `docs/overview/storage.md` | Reference events in progress section |
| `docs/reference/pantavisor-configuration.md` | `PV_LOG_EVENTS` config docs |
| `GEMINI.md` | Add event journal to overview table |


docs/overview/storage.md
anibalportero, line 279: Add events to this progress example :)

docs/overview/event-journal.md
```diff
@@ -0,0 +1,115 @@
+# Event Journal
```
anibalportero, line 1: I would move this under storage.md#update-progress.md. I would keep it a bit shorter: this intro, the overview, configuration and control.

Then I'd put there event structure and types and so on in a new reference page named pantavisor-update-progress.md with all specs for that JSON, including the events.

docs/reference/pantavisor-configuration.md
```diff
@@ -176,6 +177,7 @@ This table shows the [configuration levels](../../pantavisor-src/docs/overview/p
 | `PV_LIBTHTTP_CERTSDIR`               | :material-check: | :material-close: | :material-check: | :material-check: | :material-close: | :material-close: | :material-close: |
 | `PV_LIBTHTTP_LOG_LEVEL`              | :material-check: | :material-close: | :material-check: | :material-check: | :material-check: | :material-check: | :material-close: |
 | `PV_LOG_BUF_NITEMS`                  | :material-check: | :material-close: | :material-check: | :material-check: | :material-check: | :material-close: | :material-close: |
+| `PV_LOG_EVENTS`                      | :material-check: | :material-close: | :material-check: | :material-check: | :material-check: | :material-close: | :material-check: |
```
anibalportero, line 180: the last column should be close as that is commands. While the one left to it should be check as that is user meta.

event_log.h
```diff
@@ -0,0 +1,57 @@
...
+};
+
+struct pv_event_log {
+	struct pv_event entries[PV_EVENT_RING_SIZE];
+	int head;
+	int count;
+};
+
+void pv_event_log_init(void);
+void pv_event_log_push(pv_event_type_t type, const char *src, const char *event,
```
anibalportero, line 53: can you add variable arguments so we save some snprintf out there?

```
void pv_event_log_push(pv_event_type_t type, const char *src, const char *event, const char *detail, ...);
```

event_log.h
```diff
@@ -0,0 +1,57 @@
+/*
```
anibalportero, line 1: I would put this in its own dir... I know it's just one file, but we are doing things that way nowadays to keep it clean and in case it grows.

As event and log are things that we already have around and might be confusing, I'd prefer if we used journal/journal.h, we could then change the functions in here to pv_journal_serialize, etc.

anibalportero (commented):
I would say we should decouple this from updates. The problem with having this in progress update is that the .pv/progress is only refreshed when an update status change happens. It could be that the update finishes, but we could still be logging events to the journal after that if a container issued a reboot, for example. It is also conceptually messy.

The solution would be to attach it directly to a revision instead, keeping a .pv/journal file, serialize each time a journal entry is logged and also requires a new get in /steps ctrl endpoint. Docs would need to be adapted to this new reality.

PR #684, opened 2026-04-10
Timeout on tsh commands 

fortizc:
## Summary

- Add `tsh_run_timeout()` and `tsh_run_io_timeout()` — variants of `tsh_run()` and `tsh_run_io()` with an explicit `timeout_s` parameter; existing functions keep their signature and use the 60 s default
- Add `tsh_run_logserver_timeout()` — timeout variant of `tsh_run_logserver()`
- Replace `tsh_run_output()` pipe-based implementation with `memfd_create` to fix a deadlock: with pipes and `wait=1`, if the child output exceeded the pipe buffer (~64 KB) the child would block on write while the parent was blocked in `waitpid`; `memfd_create` gives the child an unbounded memory-backed file so writes never block
- Extract `_tsh_wait()` (signal-based wait with timeout) and `_tsh_eval_output()` (exit-status evaluation) to eliminate duplicated logic
- Harden all command parameters to `const char *`
- Fall back to `fprintf(stderr)` when the log server is unavailable (`DISABLE_LOGSERVER`)
- Fix memory leaks and logic errors found during refactor

### DIFF
--- a/utils/tsh.c
+++ b/utils/tsh.c
@@ -31,14 +31,14 @@
 #include <unistd.h>
 #include <string.h>
 #include <signal.h>
+#include <sys/mman.h>
 
 #include <sys/types.h>
 #include <sys/wait.h>
 
 #include "tsh.h"
-#include "timer.h"
 #include "utils/pvsignals.h"
-#include "log.h"
+#include "utils/fs.h"
 
 #ifndef DISABLE_LOGSERVER
 #include "logserver/logserver.h"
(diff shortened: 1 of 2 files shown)


asac:
it feels like we addd a timeout artugment to all calls event hough that want to use defaults. why keep the old function that then calls a tsh_io_run_timeout(... ) with the actual default value?

utils/tsh.c
```diff
@@ -31,31 +31,42 @@
...
+#include "utils/fs.h"
 
 #ifndef DISABLE_LOGSERVER
 #include "logserver/logserver.h"
 #endif
 
 #define TSH_MAX_LENGTH 32
 #define TSH_DELIM " \t\r\n\a"
+#define TSH_DEFAULT_TIMEOUT (60)
+#define TSH_NO_TIMEOUT (-1)
```
asac: use this in header and dont make 0 special case

PR #701, opened 2026-04-29
fix(update): rollback local revs when Hub turns unstable mid-testing

asac:
## Summary

Under `PV_CONTROL_REMOTE_ALWAYS=1`, a device with a local revision that successfully talks to the Hub and then loses Hub reachability during TESTING would silently commit to DONE. This happens because the early `pv_update_is_local() goto out` bypass in `pv_wait_network()` skips the existing `is_testing && got_any_failure` rollback branch entirely for local revs.

This change introduces a per-boot, in-memory \"Hub was reachable at some point this boot\" flag and uses it to gate a rollback during TESTING, applied **before** the local bypass.

## Behavior

| Mode | Local rev | Hub history this boot | Behavior |
|---|---|---|---|
| `REMOTE=1` only | yes | n/a | unchanged (`remote_mode=false`, `pv_wait_network` not called) |
| `REMOTE_ALWAYS=1` | yes | never online | unchanged — commits on goal achievement (no first-boot regression) |
| `REMOTE_ALWAYS=1` | yes | was online, then unstable in TESTING | **new** — rolls back |
| any | no (remote rev) | was online, then unstable in TESTING | rolls back via the new branch (also via the existing one, unchanged) |

## Implementation

- `session.was_ever_online` (bool) added to `struct pv_pantahub_session` in `pantahub_proto.c`. Initialized to false in `pv_pantahub_proto_init()`. Set to true inside `_recv_post_auth_cb()` when a session token is parsed (i.e. first successful auth). Never cleared until pantavisor restarts.
- `pv_pantahub_proto_was_ever_online()` getter exposed via `pantahub_proto.h`, wrapped as `pv_pantahub_was_ever_online()` in `pantahub.{c,h}`.
- `pv_wait_network()` in `pantavisor.c`: new branch hoisted above the local bypass, gated on `pv_pantahub_was_ever_online() && pv_pantahub_got_any_failure()` during TESTING. Existing branches preserved as a safety net.

## Test plan

- [ ] PV_CONTROL_REMOTE_ALWAYS=1, factory-style local revision, no network ever — boots, reaches DONE, commits.
- [ ] PV_CONTROL_REMOTE_ALWAYS=1, local revision, brought online (auth succeeds) then network cut during TESTING — observe \"Hub was reachable earlier this boot but became unstable during testing. Rolling back...\" and a rollback.
- [ ] PV_CONTROL_REMOTE_ALWAYS=1, local revision, fully online during TESTING — commits to DONE as before.
- [ ] PV_CONTROL_REMOTE=1 only, local revision — unchanged path, commits regardless of Hub state.
- [ ] Remote (Hub-pulled) revision — existing INPROGRESS-without-reporting and TESTING-with-failure paths still trigger rollback (regression).

### DIFF
--- a/pantahub/pantahub.c
+++ b/pantahub/pantahub.c
@@ -952,6 +952,11 @@ bool pv_pantahub_is_online()
 	return pv_pantahub_proto_is_online();
 }
 
+bool pv_pantahub_was_ever_online()
+{
+	return pv_pantahub_proto_was_ever_online();
+}
+
 bool pv_pantahub_got_any_failure()
 {
 	return pv_pantahub_proto_got_any_failure();
--- a/pantahub/pantahub.h
+++ b/pantahub/pantahub.h
@@ -62,6 +62,7 @@ void pv_pantahub_evaluate_state(void);
 bool pv_pantahub_is_reporting(void);
 
 bool pv_pantahub_is_online(void);
+bool pv_pantahub_was_ever_online(void);
 bool pv_pantahub_got_any_failure(void);
 
 bool pv_pantahub_is_progress_queue_empty(void);
--- a/pantahub/pantahub_proto.c
+++ b/pantahub/pantahub_proto.c
@@ -76,6 +76,7 @@ typedef enum {
 struct pv_pantahub_session {
 	char *token;
 	bool online;
+	bool was_ever_online;
 	short failed_requests;
 	bool any_failed_request;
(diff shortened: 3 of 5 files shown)


asac:
this PR creates issues in no-reboot update scenario... it will believe there is a pantahub outage .... not sure why ...needs fixing.

PR #703, opened 2026-05-05
Remove log about the object progress

fortizc:
This PR removes the progress lo to avoid unnecessary noise on pantavisor log


asac:
lgtm

PR #712, opened 2026-05-15
feat(logserver): add RFC 3164 and RFC 5424 syslog protocol support

fortizc:
Restructure logserver protocol handling under `logserver/proto/` with a
unified interface for both binary and syslog-based protocols.

## Changes

- Move the binary protocol to its own module (`logserver_binary`)
- Add `logserver_rfc` with shared syslog utilities: priority-to-level
  mapping, facility extraction, `/dev/log` symlink creation, and
  RFC 3164 vs 5424 auto-detection
- Add `logserver_rfc3164` for BSD syslog parsing (timestamp, host,
  app-name with PID bracket stripping)
- Add `logserver_rfc5424` for IETF syslog parsing with nil-value
  handling and structured-data skipping


anibalportero:
Docs should be changed to present this new feature.

asac:
whats the status?

fortizc:
> whats the status?

Currently almost everything works, the parser, message are received correctly, /dev/log is mounted from the host using lxc. The only problem is how to determine the name of the container. 
To use cgroups lib I need a PID, the available PID comes from the application which is sending logs, so I need the parent PID to get the right name. The issue with this approach is when the app logs a message and then finish (app is closed), sometimes I logserver cannot get the parent PID or even its own PID

anibalportero:
lgtm

plugins/pv_lxc.c
```diff
@@ -832,6 +832,12 @@ int pv_start_container(struct pv_platform *p, const char *rev, char *conf_file,
 		if (p->exec)
 			c->set_config_item(c, "lxc.init.cmd", p->exec);
 
+		if (!c->set_config_item(
+			    c, "lxc.mount.entry",
+			    "/dev/log dev/log none bind,create=file 0 0")) {
+			pv_log(WARN, "/dev/log not mounting");
+		}
+
```
asac, line 843: lets make a config in run.json to enable/disable this logger variant. and have a default config flag in pantavisor.config that woul denable this logger by default on all containers.

docs/reference/logserver-sockets.md
```diff
@@ -64,3 +64,165 @@ The `msghdr` must contain an `iovec` array with 4 elements:
 
 !!! Note
     Only one file descriptor can be subscribed per platform-source pair. Subscribing a new FD for an existing pair will replace the previous one.
+
+## /dev/log
+
```
asac, line 69: here we should reference what config convenience we offer to inject this automatically.

asac: also i think the docs in general about the configs are missing in this PR.

fortizc: docs updated

plugins/pv_lxc.c
asac: what about the per-container config in run.json?

fortizc: the last commit adds "standard_log" to the run.json this key controls if the /dev/log is mounted or not on that container. 
Also docs are updated

docs/reference/pantavisor-state-format-v2.md
```diff
@@ -135,6 +135,7 @@ Configures an individual container runtime.
 | `drivers` | object | No | Requirements: `required`, `optional`, or `manual`. |
 | `services` | object | No | [Service mesh requirements](#service-requirements). |
 | `logs` | array | No | [Logger configurations](#loggers-array). |
+| `standard_log` | boolean | No | Whether to bind-mount `/dev/log` into this container. Defaults to the global [`PV_LOG_STD_SUPPORT`](pantavisor-configuration.md#summary) setting. Has no effect when `PV_LOG_STD_SUPPORT` is globally disabled. |
```
asac: imo this config approach is a bit awkward.

Right now you cannot disable stdsupport globally and then enable on a per container basis. At best our config approach would allow for both (default off, enable specific as well as default on and disable specifics).

Also i dont see why we would disable the support of the protocol on that socket. I think we can achieve all by rather making the auto dev/log injection logic switch, like global PV_LOG_AUTO_DEVLOG would enable/disable the default dev/log injection. and then have the ability at a container level to do auto_devlog true and false for the fine grained level.

PR #713, opened 2026-05-15
perf(event): add libevent priorities to prevent ctrl starvation during transfers

anibalportero:
## Summary

Pantavisor uses a single libevent event base shared by the ctrl Unix socket server, outgoing hub connections, and all file transfers. With a single priority level, a continuous bulk transfer (hub artifact download or ctrl `PUT/GET /objects`) starves ctrl requests: the event loop keeps servicing the transfer's bufferevent/timer callbacks, leaving pvcontrol queries unanswered for tens of seconds — long enough to cause pvtest `update-retries-pv-crash` to observe a state that has already transitioned past the expected value.

This PR introduces 4 libevent priority levels so each category of I/O yields to more-latency-sensitive categories within the same dispatch cycle.

## Priority levels (`event/event.h`)

| constant | value | assigned to |
|---|---|---|
| `PV_EVENT_PRIORITY_HIGH` | 0 | reserved, unused |
| `PV_EVENT_PRIORITY_DEFAULT` | 1 | state machine one-shots, hub API calls |
| `PV_EVENT_PRIORITY_CTRL` | 2 | ctrl non-file-transfer requests (libevent default with COUNT=4) |
| `PV_EVENT_PRIORITY_LOW` | 3 | hub artifact downloads, ctrl `PUT/GET /objects` |

## Changes

- `event/event.h` — define the four `PV_EVENT_PRIORITY_*` constants and `PV_EVENT_PRIORITY_COUNT`
- `event/event.c` — call `event_base_priority_init(base, PV_EVENT_PRIORITY_COUNT)` immediately after `event_base_new()`
- `event/event_rest.h` + `event/event_rest.c` — add `int priority` parameter to `pv_event_rest_send_by_components` and `pv_event_rest_send_by_url`; call `bufferevent_priority_set(bev, priority)` on the outgoing connection
- `pantahub/pantahub_proto.c` — pass `PV_EVENT_PRIORITY_DEFAULT` for hub API calls (`_send_by_endpoint`), `PV_EVENT_PRIORITY_LOW` for hub artifact downloads (`_get_object`)
- `ctrl/ctrl_download.c` — set `PV_EVENT_PRIORITY_LOW` on the chunk-send timer so ctrl object downloads yield to regular ctrl requests
- `ctrl/ctrl_objects_ep.c` — demote the connection's bufferevent to `PV_EVENT_PRIORITY_LOW` at the start of `PUT /objects` upload and `GET /objects` download handlers

## Open questions

- `pv_event_one_shot` and `pv_event_timeout` use `event_base_once`, which does not expose priority. These state machine callbacks land at the libevent default (CTRL=2) instead of DEFAULT=1. Should we refactor to `event_new` + `event_priority_set` + `event_add` to properly place them at level 1?
- Is CTRL (2) the right level for the ctrl server, or should short ctrl requests such as `/steps` be at DEFAULT (1) to ensure they are never delayed by other ctrl connections?


asac:
lets discuss... i feel there is something else wrong if we have troubles with starving because of parallel download...

PR #716, opened 2026-05-20
fix(updater): yield event loop between evbuffer writes during object download

anibalportero:
## Summary

- Replace the synchronous `evbuffer_write()` loop in the object download callbacks with a deferred-write pattern that yields to the event loop between each write
- Eliminates the ctrl socket starvation (pvcontrol timeouts) that occurred when large objects arrived in a single `chunk_cb` call on fast networks
- Data is moved from the HTTP receive buffer into an owned evbuffer via `evbuffer_add_buffer()` (O(1), no copy), then drained 256 KB at a time via `evbuffer_write_atmost()` per event loop tick using `event_base_once()` zero-timeout events

## Root cause

`chunk_cb` previously drained the entire incoming buffer to disk in a tight `evbuffer_write()` loop, blocking the libevent event loop for its full duration. On fast networks (e.g. GHA runners) the entire object (52 MB in tests) arrives in a single `chunk_cb` invocation, producing a multi-second block during which ctrl requests cannot be dispatched.

On slow networks the data trickles in across many small `chunk_cb` calls, each draining in one syscall, so the gap was masked locally.

## Changes

- Add `struct pv_object_writer` to carry the evbuffer, fd, path, id_ref, HTTP status, and finalization state across deferred callbacks
- `_get_object()`: allocate writer, open tmp file, pass writer as callback context
- `_recv_get_object_chunk_cb()`: `evbuffer_add_buffer()` into writer's buffer, schedule write if not already pending
- `_object_writer_write_cb()`: one `evbuffer_write_atmost()` call (256 KB cap), re-schedule if data remains, call finish when buffer empty and finalized
- `_recv_get_object_done_cb()`: save HTTP status before req is freed, drain remaining input bytes, set finalized flag
- `_object_writer_finish()`: close fd, remove transfer, install object, evaluate state

### DIFF
--- a/pantahub/pantahub_proto.c
+++ b/pantahub/pantahub_proto.c
@@ -20,16 +20,20 @@
  * SOFTWARE.
  */
 
+#include <errno.h>
 #include <fcntl.h>
 #include <limits.h>
 #include <string.h>
 #include <unistd.h>
+
+#include <event2/buffer.h>
 
 #include "pantahub/pantahub.h"
 #include "pantahub/pantahub_msg.h"
 #include "pantahub/pantahub_proto.h"
 #include "pantahub/pantahub_struct.h"
 
+#include "event/event.h"
 #include "event/event_rest.h"
 
 #include "config.h"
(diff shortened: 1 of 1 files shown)


anibalportero:
finally going with https://github.com/pantavisor/pantavisor/pull/718 solution

PR #718, opened 2026-05-20
fix(pantahub): cap download read buffer to prevent event loop starvation

anibalportero:
## Summary
- During large binary downloads, libevent read all available TCP data into
  the bufferevent in one shot, causing chunk_cb to fire with the entire
  payload (up to 50 MB) at once
- The synchronous write loop blocked the event loop for the full download
  duration, starving pvcontrol requests ("ctrl gap")
- A 128 KB read high-watermark limits data buffered per event-loop tick;
  combined with BEV_OPT_DEFER_CALLBACKS (already set), ctrl requests are
  dispatched between download chunks

## Changes
- `event/event_rest.c`: add `HTTP_DOWNLOAD_READ_HIGHWATER` (128 KB) constant
- `event/event_rest.c`: call `bufferevent_setwatermark(bev, EV_READ, 0, HTTP_DOWNLOAD_READ_HIGHWATER)` after creating the TLS bufferevent in `pv_event_rest_send_by_components()`

### DIFF
--- a/event/event_rest.c
+++ b/event/event_rest.c
@@ -48,6 +48,7 @@
 #include "utils/str.h"
 
 #define MODULE_NAME "event_rest"
+#define HTTP_DOWNLOAD_READ_HIGHWATER (128 * 1024)
 #define pv_log(level, msg, ...)                                                \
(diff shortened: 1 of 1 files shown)


asac:
were we able to still see the issue with this one?

PR #721, opened 2026-05-25
fix(pvcontrol): prefer pvcurl over curl

anibalportero:
## Summary

Swap backend selection order in `pvcontrol`: try `pvcurl` first, fall back to `curl`. `pvcurl` is always present on Pantavisor devices while system `curl` may not be installed. Remove the experimental warning that was emitted when pvcurl was selected.

## Test plan

- [x] `local/control/basic-endpoints-curl` — pvcontrol with forced `CURL_CMD=curl` still works. **PASSED**
- [x] `local/control/basic-endpoints` — pvcontrol uses pvcurl by default, output is correct. **PASSED**
- [ ] `remote/control/device-user-metadata` — requires pantavisor/pantavisor#722 (ctrl evbuffer guard) to be merged first; the raw HTTP response in that test shows the double Content-Type header which is fixed there, not here.

### DIFF
--- a/tools/pvcontrol
+++ b/tools/pvcontrol
@@ -10,19 +10,15 @@ response=''
 signal=''
 daemon_name=''
 
-CURL_CMD=${CURL_CMD:-curl}
+CURL_CMD=${CURL_CMD:-pvcurl}
 
 if ! command -v $CURL_CMD >/dev/null 2>&1; then
-	CURL_CMD="pvcurl"
+	CURL_CMD="curl"
 fi
 
 if ! command -v $CURL_CMD >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


tools/pvcurl
```diff
@@ -125,7 +125,7 @@ if [ -n "$OUTPUT_FILE" ]; then
 elif [ "$SHOW_HEADERS" = true ]; then
     cat /tmp/http_res
 else
-    sed '1,/^\r\{0,1\}$/d' /tmp/http_res
+    extract_body
```
asac, line 128: do you know what the previous one did and why it didnt use extract_body?

PR #722, opened 2026-05-26
fix(ctrl): guard all evbuffer callbacks against re-entrant drain

anibalportero:
## Summary

`evbuffer_add_cb` callbacks on an `evhttp` input buffer fire synchronously on both additions and drains. Without a guard the drain inside the callback re-enters it with `n_added==0`, triggering a second `evhttp_send_reply` — use-after-free / assertion crash exploitable by any local client.

**Commit 1** — adds `if (!info->n_added) return` to four remaining unguarded callbacks: `ctrl_step_commit_cb`, `ctrl_daemons_process_action`, `ctrl_utils_drain_ok_callback`, `ctrl_utils_drain_error_callback`. Places `(void)buf` before the guard per project convention.

**Commit 2** — fixes a dangling-pointer in `ctrl_utils_drain_data`: `msg` stored a pointer to the callers stack-allocated `err[]` buffer, gone before the async callback fires. Embeds the string as `char msg[PV_CTRL_MAX_ERR]`, copies it at registration time. Also frees the struct after use.

Note: equivalent guards for `ctrl_usrmeta_set_cb` and `ctrl_devmeta_set_key` were applied to master directly in `4f3ed21`.

## Test plan

- [ ] `local/control/basic-endpoints` and `local/control/basic-endpoints-curl` pass.
- [ ] `remote/control/device-user-metadata` — PUT usrmeta/devmeta produces single Content-Type header.

### DIFF
--- a/ctrl/ctrl_daemons_ep.c
+++ b/ctrl/ctrl_daemons_ep.c
@@ -68,6 +68,11 @@ static void ctrl_daemons_process_action(struct evbuffer *buf,
 					const struct evbuffer_cb_info *info,
 					void *ctx)
 {
+	(void)buf;
+
+	if (!info->n_added)
+		return;
+
 	struct evhttp_request *req = ctx;
 	char *data = NULL;
 	char *action = NULL;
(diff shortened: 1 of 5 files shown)


anibalportero:
done in !724

PR #740, opened 2026-06-26
fix(config,metadata): honor user-metadata deletions

anibalportero:
Treat an empty or `null` user-metadata value from the Hub as a **deletion**, and revert a config value that user metadata had overridden back to its baseline when the key is removed. Without this, clearing a usermeta key left the override in place — and an empty value coerced numeric config keys to `0` — so the device-metadata pvtests behaved non-deterministically.

## What changed

- **config.c / config.h** — on the first `META`-level override of a config entry, snapshot its prior value and level (`base_value`, `base_modified`, `meta_saved`). New `pv_config_unset_value()` restores that snapshot; `pv_config_free()` frees the snapshot string.
- **metadata.c** — `pv_usermeta_parse()` treats a `null` primitive or an empty string as a deletion: it calls `pv_metadata_rm_usermeta()` instead of storing an empty value. `pv_metadata_rm_usermeta()` now calls `pv_config_unset_value()` so any config override the key held is reverted.

Split out of #727 so it can land on its own; #727 rebases on top once this merges.

### DIFF
--- a/config.c
+++ b/config.c
@@ -126,6 +126,16 @@ struct pv_config_entry {
 		int i;
 		char *s;
 	} value;
+	// snapshot of value/level as it was right before the first META
+	// (user metadata) override, so the override can be reverted when the
+	// usermeta key is deleted.
+	bool meta_saved;
+	level_t base_modified;
+	struct {
+		bool b;
+		int i;
+		char *s;
+	} base_value;
 };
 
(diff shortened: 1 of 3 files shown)


anibalportero:
dup https://github.com/pantavisor/pantavisor/pull/741

PR #747, opened 2026-07-08
docs: dup CLAUDE.md instead of using reference to GEMINI.md

anibalportero:


fortizc:
I arrived a little late, but isn't better just replace both GEMINI.md and CLAUDE.md with an AGENTS.md ?

anibalportero:
> I arrived a little late, but isn't better just replace both GEMINI.md and CLAUDE.md with an AGENTS.md ?

Maybe. I was just making this equivalent to what we have in meta-pantavisor, but feel free to make any changes. I don't have a strong opinion on this.

PR #749, opened 2026-07-08
feat(pvtx): support explicit commit revision name

anibalportero:
## Summary
- Adds an optional `[name]` argument to `pvtx commit`, letting the caller supply the revision name verbatim instead of always auto-generating `locals/pvtx-<timestamp>-<hash>-<rand>`
- Fixes pv-ctrl socket lookup to also try `/run/pantavisor/pv/pv-ctrl` first, matching pvcontrol's search order

## Changes
- `pv_pvtx_txn_commit()` now takes a `name` param; uses it verbatim if non-empty, else falls back to auto-generated name
- `cmd_commit` parses optional `argv[2]` as the name
- Updated `commit` help text to document `[name]`
- `connect_sock()` tries `/run/pantavisor/pv/pv-ctrl` before `/pv/pv-ctrl` and `/pantavisor/pv-ctrl`; error message updated accordingly


asac:
didnt we add a test framework for this already? maybe wire this up there...

PR #750, opened 2026-07-10
fix(update): report download progress on the mainloop heartbeat, not per object

asac:
## Problem

Pantavisor reports OTA download progress to Pantacor Hub (and the local trail-dir
`.pv/progress` file) via the step `progress` JSON. In the Hub UI the download bar
barely moves — it updates only 2–3 times over an entire multi-minute download,
then jumps to done.

**Root cause:** byte-accounting was coupled to *whole-object completion*, not to
bytes arriving. `_recv_get_object_chunk_cb` wrote each chunk to disk but reported
nothing; only `_recv_get_object_done_cb` (once per object) added the whole object
size via `pv_update_progress_add_downloaded(o->size)`. A pantavisor update is
dominated by a few large objects, so a big object downloading for minutes produced
**zero** intermediate updates.

## Change: decouple accounting from reporting

- **Accounting** (per chunk) is a pure in-memory increment — no serialize, no
  disk, no PUT on the hot path. Bytes are tracked **per transfer** in
  `pantahub_proto` and **rolled back** if an object fails/retries, so the counter
  can never exceed `total_size`.
- **Reporting** (serialize → trail-dir file → coalesced Hub PUT) is driven by the
  existing 6s download heartbeat (`REQ_INTERVAL`) in `_download_objects_cb` via
  `pv_update_report_progress()`, and is skipped when no bytes moved. Status-change
  reports still reload logs; the byte cadence reuses the in-memory copy.
- The whole-object add in `pv_update_install_object` is **removed** — per-chunk
  received bytes already sum to the object size, so keeping it would double count.
- `event_rest` recv helpers gain a `size_t *written` out-param so callers can feed
  arriving bytes into accounting.

No new config key or on-disk structure: `downloads.total.total_downloaded` already
rides in the one shared progress blob written to both the trail-dir file and the Hub.

## Result

`total_downloaded` now climbs smoothly (~every 6s) through even a single large
object, and is accurate across retries.

**Validated on device (opi3_hyena):** a 73.6 MB multi-object update reported
continuous progress instead of jumping only at object boundaries:

```
dl=0 → 20.8MB → 42.7MB → 53.9MB → 63.7MB → 71.6MB  (of 73.6MB), ~6–7s apart → DONE
```

## Touched files

`event/event_rest.{c,h}`, `pantahub/pantahub.c`, `pantahub/pantahub_proto.c`,
`update/update.{c,h}`, `update/update_progress.{c,h}`, `update/update_struct.h`

### DIFF
--- a/pantahub/pantahub_proto.c
+++ b/pantahub/pantahub_proto.c
@@ -64,6 +64,7 @@ struct pv_progress {
 struct pv_object_transfer {
 	const char *id_ref;
 	bool active;
+	off_t received; // bytes received so far for this object
 	struct dl_list list; // struct pv_object_transfer
 };
 
@@ -756,40 +757,62 @@ int pv_pantahub_proto_get_objects_metadata()
 static void _recv_get_object_chunk_cb(struct evhttp_request *req, void *ctx)
 {
 	char path[PATH_MAX];
-	int res;
+	size_t written = 0;
 	const char *id_ref = (char *)ctx;
+	struct pv_object_transfer *o;
 
 	pv_log(TRACE, "run event: cb=%p", (void *)_recv_get_object_chunk_cb);
 
 	pv_storage_set_object_download_path(path, PATH_MAX, id_ref);
-	pv_event_rest_recv_chunk_path(req, path);
+	if (pv_event_rest_recv_chunk_path(req, path, &written))
+		return;
+
+	o = _search_object_transfer(id_ref);
+	if (o)
+		o->received += written;
+	pv_update_add_downloaded((off_t)written);
 }
 
 static void _recv_get_object_done_cb(struct evhttp_request *req, void *ctx)
 {
 	char path[PATH_MAX];
+	size_t written = 0;
 	int res;
 	const char *id_ref = (char *)ctx;
+	struct pv_object_transfer *o;
+	off_t received = 0;
 
 	pv_log(TRACE, "run event: cb=%p", (void *)_recv_get_object_done_cb);
 
+	pv_storage_set_object_download_path(path, PATH_MAX, id_ref);
+	res = pv_event_rest_recv_done_path(req, path, &written);
+
+	// account the final flush; read received before _remove frees the transfer
+	o = _search_object_transfer(id_ref);
+	if (o) {
+		o->received += written;
+		received = o->received;
+	}
+	pv_update_add_downloaded((off_t)written);
+
 	_remove_object_transfer(id_ref);
 
-	pv_storage_set_object_download_path(path, PATH_MAX, id_ref);
-	res = pv_event_rest_recv_done_path(req, path);
 	if (res == 401) {
-		pv_log(WARN, "GET object unauthorized", res);
+		pv_log(WARN, "GET object unauthorized");
 		_free_token();
+		pv_update_add_downloaded(-received); // roll back partial
 		goto out;
 	}
 	if (res != 200) {
 		pv_log(WARN, "GET object returned %d", res);
+		pv_update_add_downloaded(-received); // roll back partial
 		goto out;
 	}
 	pv_log(DEBUG, "object downloaded from Hub");
 
 	if (pv_update_install_object(path)) {
 		pv_log(WARN, "object download failed");
+		pv_update_add_downloaded(-received); // roll back partial
 		goto out;
 	}
 out:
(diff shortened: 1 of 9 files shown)


update/update.c
```diff
@@ -502,6 +504,24 @@ int pv_update_install_object(const char *in_path)
...
+	pv_update_progress_add_downloaded(&u->progress, downloaded);
+}
+
+void pv_update_report_progress(void)
+{
+	struct pv_update *u = _get_update_instance();
+	if (!u)
+		return;
+
+	pv_update_progress_report(&u->progress);
```
asac, line 518: having names pv_update_report_progress and pv_update_progress_report is confusing. one of them could be more specific i guess... do we need both funcs in exported api? maybe one is static?

update/update.c
```diff
@@ -485,7 +485,9 @@ int pv_update_install_object(const char *in_path)
 		goto out;
 	}
 
-	pv_update_progress_add_downloaded(&u->progress, o->size);
+	// download bytes are now accounted per chunk in pantahub_proto.c; the
+	// per-transfer received bytes already sum to the object size, so adding
+	// o->size here would double count
```
asac: we dont need this comment. rather explain that in the commit message.

pantahub/pantahub_proto.c
```diff
@@ -756,40 +757,63 @@ int pv_pantahub_proto_get_objects_metadata()
...
 	if (res != 200) {
 		pv_log(WARN, "GET object returned %d", res);
+		pv_update_add_downloaded(-received); // roll back partial
 		goto out;
 	}
 	pv_log(DEBUG, "object downloaded from Hub");
 
 	if (pv_update_install_object(path)) {
 		pv_log(WARN, "object download failed");
+		pv_update_add_downloaded(-received); // roll back partial
```
asac, line 815: do we need these for the retry logic to have correct accounting? otherwise i guess keeping the total downlaoded even if failed in memory makes sense.

pantahub/pantahub_proto.c
```diff
@@ -756,40 +757,63 @@ int pv_pantahub_proto_get_objects_metadata()
...
+	struct pv_object_transfer *o;
+	off_t received = 0;
 
 	pv_log(TRACE, "run event: cb=%p", (void *)_recv_get_object_done_cb);
 
+	pv_storage_set_object_download_path(path, PATH_MAX, id_ref);
+	res = pv_event_rest_recv_done_path(req, path, &written);
+
+	// account the final flush, then capture received *before* removing the
+	// transfer (removal frees it)
```
asac: if a comment just state that this is the accounting for final flush.... keep it short

update/update.c
asac: Agreed the near-anagram is confusing. Renamed the instance-level entry point to `pv_update_report_download_progress()` (what pantahub calls on the download heartbeat), keeping the generic struct-level primitive `pv_update_progress_report(struct pv_update_progress *)`.

We do want both — it's the same two-layer split as `pv_update_add_downloaded()` → `pv_update_progress_add_downloaded()`: the update.c wrapper resolves the update singleton so pantahub never reaches into `struct pv_update`, and the update_progress.c function does the accounting/skip-if-unchanged logic. The latter can't be static since update.c calls it across translation units. Done in bac7b755.

update/update.c
asac: Agreed — removed. The double-count rationale is already in the commit message body. bac7b755.

pantahub/pantahub_proto.c
asac: Agreed — shortened to `// account the final flush; read received before _remove frees the transfer`. Kept the one-line note because reading `o->received` after `_remove_object_transfer()` frees `o` would be a use-after-free. bac7b755.

pantahub/pantahub_proto.c
asac: Yes — the rollback is needed for correct accounting across retries. On a failed object we `pv_fs_path_remove(path)` and the object goes back to unavailable, so `pv_pantahub_proto_get_objects()` re-adds it and re-downloads the whole object from scratch.

The failed attempt's chunks were already added to `total.downloaded` per-chunk (in `_recv_get_object_chunk_cb`). Without the `-received` rollback, the retry would add the full object size again → `total_downloaded` overshoots `total_size` and the Hub bar goes past 100%. So rather than 'keep failed bytes in memory', we back out exactly this transfer's bytes and let the retry re-account cleanly. This is the over-count case the design guards against, so I'd keep it as-is.

PR #761, opened 2026-07-16
feat(ctrl): add synchronous /storage/gc and unclaim commands

anibalportero:
## Summary
Two small ctrl API additions, pulled out of the pvtest-device-target work where they're used, since they're generic and independently useful.

## Changes
- `POST /storage/gc`: synchronous garbage-collect endpoint (`ctrl_storage_ep.c`), plus `pvcontrol storage gc`
- `cmd unclaim`: strips Hub credentials from the live config so the device walks idle→register→claim on its own, no reboot required; plus `pvcontrol cmd unclaim`


anibalportero:
Split into separate PRs: #764 (storage gc) and #763 (unclaim).

PR #774, opened 2026-07-24
feat(logserver): send lxc logs to stdout_direct when active

anibalportero:


platforms.c
```diff
@@ -67,9 +67,6 @@
 	     ##__VA_ARGS__)
 #include "log.h"
 
-#define PV_PLATFORM_LXC_LOG "lxc/lxc.log"
-#define PV_PLATFORM_LXC_CONSOLE_LOG "lxc/console.log"
-
```
fortizc, line 72: I'm not sure about this, removes from here and add it to logserver.h makes this module very dependent on logserver. 
I feel this should stay in platform and be used by logserver

PR #779, opened 2026-07-30
fix(logserver): avoid losing messages from short-lived processes.

fortizc:
## General description
With the new protocols in Logserver (specifically RFC3164 and RFC5424), we need to retrieve the process cgroup to route log messages to the correct destination file. Currently, this is done by querying socket APIs to get the PID and reading `/proc/<PID>/cgroup`. 

However, this creates an issue for short-lived processes: if a process terminates before we read `/proc/<PID>/cgroup`, its cgroup can no longer be retrieved.

## Proposal
This PR addresses the issue by capturing connection details immediately when a connection is accepted. 
To achieve this, the internal Logserver structure will maintain a list of connection metadata, including:
* The current file descriptor
* The originating process ID 
* The associated cgroup

Once the connection/FD is fully processed (i.e., the log message or command has been read), its entry is removed from the list.

### DIFF
--- a/logserver/logserver.c
+++ b/logserver/logserver.c
@@ -90,6 +90,13 @@ struct logserver_fd {
 	struct dl_list list;
 };
 
+struct logserver_conninfo {
+	int fd;
+	pid_t pid;
+	char *cgroup;
+	struct dl_list list;
+};
+
 struct logserver {
 	pid_t pid;
 	pid_t cmd_pid;
@@ -108,6 +115,7 @@ struct logserver {
 	// only if was sent to the fd_sock
 	struct dl_list tmplst;
 	struct dl_list outputs;
+	struct dl_list conninfo;
 };
 
 static struct logserver logserver = {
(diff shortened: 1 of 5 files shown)


logserver/logserver.c
```diff
@@ -546,39 +653,14 @@ static void logserver_remove_fd(int fd)
...
-		buffer->buf[end] = '\0';
-
-		pid_t sender_pid = pv_socket_get_sender_pid(fd);
-		logserver_handle_msg(buffer->buf, sender_pid);
-	} else if (errno != EAGAIN) {
-		pv_log(DEBUG, "dead fd (%d) found trying to read: %s", errno,
-		       strerror(errno));
-		logserver_remove_fd(fd);
+	struct logserver_conninfo *ci = logserver_conninfo_search(fd);
+	if (ci) {
```
anibalportero, line 660: is this the only exit point? shouldn't we clean this before logserver loop ends?

logserver/logserver.c
```diff
@@ -708,6 +790,12 @@ static void logserver_loop()
 				continue;
 			}
 
+			struct logserver_conninfo *ci =
```
anibalportero, line 796: is it possible that whatever opens the communication dies before accept? or, if by any chance, could we return NULL? if so, how do we fare? are we still logging as "unknown-platform" in any case?

anibalportero:
It would be nice to exercise this a bit in pvtest, maybe in local/services. 

Maybe it would be difficult to add the program that deterministically closes before logserver processing, but at least one that covers the main path would be great. I suppose you would need to add a new example container for this.

logserver/logserver.c
fortizc: Yes this is the only exit point. This is because the loop always calls this function for used fd or if found any error in the fd. So always finish at this point similar to `logserver_fd`

logserver/logserver.c
fortizc: Ok, good catch, now I introduce a small fix to always logs "unknown-platform" for NULL cgroups.
The chances to get errors exists, we still could lose connection before get the information, but the only 100% reliable way is to introduce a eBPF program and use it here. While the program itself should be simple, the integration could be a bit more complex, maybe too much for this simple problem.

PR #782, opened 2026-08-05
fix(cgroup): pantavisor cgroup name collision with other appengine instances

anibalportero:
Appengine instances share the host cgroupfs, so a fixed name resolves to the same directory for every instance running at once: they collide on identical container names and remove each other's leaves on destroy.

Suffix the name with the cgroup we are started in, which is unique per appengine container and stable across its restarts.

### DIFF
--- a/cgroup.c
+++ b/cgroup.c
@@ -102,6 +102,59 @@ static int pv_cgroup_mkcgroup_root()
 	return ret;
 }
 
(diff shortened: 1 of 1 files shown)


asac:
trying to do it more right here: https://github.com/pantavisor/pantavisor/pull/796

PR #790, opened 2026-08-11
Add configurable alerts for platform status changes

fortizc:
This PR adds a new `WALL` log level for events that should be seen on the device console while they happen, and uses it for container status changes.

`WALL` is appended after `TRACE` in `enum log_level`, so the numeric values of the existing levels (used by `PV_LOG_LEVEL`, `PV_LIBTHTTP_LOG_LEVEL` and the binary log protocol) do not change. For filtering, `WALL` has the same precedence as `INFO`: a new helper, `pv_log_level_rank()`, maps `WALL` to `INFO` and is used everywhere a
level is compared against `PV_LOG_LEVEL`.

Pantavisor writes every `WALL`, `ERROR` and `FATAL` message it logs to the console. This happens even when `PV_LOG_LEVEL` filters the message out of the logs, so alerts reach the console with the default log
level. The console line is skipped when a configured stdout output already shows the message, to avoid duplicates. Container logs are not mirrored.

To support this, the `wall` module gets new functions to send formatted messages to the console:

```c
void pv_wall_v(const char *message, va_list args);
void pv_wall_vlog(int level, const char *src, const char *message, va_list args);
void pv_wall_log(int level, const char *src, const char *message, ...);
```

The arguments are:

* level: log level of the message, printed as its name.
* src: the source, usually the module name (`MODULE_NAME` in most Pantavisor modules).
* message: the message itself, supports all the `printf()` format specifiers (%s, %c, etc).
* args: arguments for the format specifiers, if any.

Each line is prefixed with the time since boot (CLOCK_MONOTONIC), like dmesg. The line is built in a 1024 bytes buffer; longer messages are truncated, and an ellipsis is appended.

Console output can be disabled with the new `PV_LOG_CONSOLE_ALERTS` configuration key (default `1`). The key is only read at boot: it can be set in pantavisor.config, OEM config, env/bootargs or policies, but not through user metadata or commands. Console alerts are documented in the overview and reference.

pvlogger no longer reports an error when `pv_logserver_send_log()` returns 0: that value means the message was filtered by `PV_LOG_LEVEL`, and the error report recursed into itself.

Example output for container status changes:

```
[    6.217276] [PANTAVISOR] [platforms] WALL: platform 'awconnect' status is now STARTING
[    6.217450] [PANTAVISOR] [platforms] WALL: platform 'pvr-sdk' status is now BLOCKED
[    6.341446] [PANTAVISOR] [platforms] WALL: platform 'awconnect' status is now STARTED
[    6.419898] [PANTAVISOR] [platforms] WALL: platform 'pvr-sdk' statusis now MOUNTED
[    6.480485] [PANTAVISOR] [platforms] WALL: platform 'pvr-sdk' statusis now STARTING
[    6.653835] [PANTAVISOR] [platforms] WALL: platform 'pvr-sdk' status is now STARTED
```

### DIFF
--- a/config.c
+++ b/config.c
@@ -26,6 +26,7 @@
 #include <unistd.h>
 #include <stdio.h>
 #include <errno.h>
+#include <ctype.h>
 
 #include <sys/stat.h>
 #include <mntent.h>
@@ -44,6 +45,7 @@
 #include "paths.h"
 #include "pantavisor.h"
 #include "parser/parser.h"
+#include "platforms.h"
 #include "utils/fs.h"
 #include "utils/str.h"
 #include "utils/math.h"
@@ -67,7 +69,8 @@ typedef enum {
 	SB_MODE,
 	STR,
 	TIMER_TYPE,
-	WDT_MODE
+	WDT_MODE,
+	PLATFORM_STATUS
 } type_t;
 
 typedef enum {
@@ -241,6 +244,8 @@ static struct pv_config_entry entries[] = {
 	{ STR, "PV_NET_BRMASK4", PV | OEM, 0, false,
 	  .value.s = NET_BRMASK4_DEF },
 	{ STR, "PV_OEM_NAME", PV, 0, false, .value.s = NULL },
+	{ PLATFORM_STATUS, "PV_PLATFORM_ALERT_STATUS", PV | OEM, 0, false,
+	  .value.i = 0 },
 	{ STR, "PV_POLICY", PV, 0, false, .value.s = NULL },
 	{ PWR_MODE, "PV_POWER_MODE", PV | OEM | RUN, 0, false,
 	  .value.i = PWR_LOCKS },
@@ -604,6 +609,87 @@ _set_config_by_entry_log_server_outputs(struct pv_config_entry *entry,
 
 	server_outputs |= LOG_SERVER_OUTPUT_UPDATE;
 	entry->value.i = server_outputs;
+}
+
+static void _set_config_by_entry_platform_status(struct pv_config_entry *entry,
+						 const char *value)
+{
+	if (!entry)
+		return;
+
+	char *tmp = NULL;
+	char *val = strdup(value);
+	if (!val)
+		return;
+
+	int status_code = 0;
+
+	for (char *str = val;; str = NULL) {
+		char *tok = strtok_r(str, ",", &tmp);
+		if (!tok)
+			break;
+
+		int len = strlen(tok);
+		for (int i = 0; i < len; i++)
+			tok[i] = toupper((unsigned char)tok[i]);
+
+		if (pv_str_matches(tok, len, "ALL", strlen("ALL"))) {
+			status_code = 0xffff;
+			goto out;
+		}
+
+		int status = pv_platform_status_from_string(tok);
+		if (status == -1) {
+			pv_log(DEBUG,
+			       "Unrecognized plataform status %s changes will "
+			       "be no reported",
+			       tok);
+			continue;
+		}
+
+		status_code |= 1 << status;
+	}
+
+out:
+	if (val)
+		free(val);
+
+	entry->value.i = status_code;
+}
+
+static char *_get_platform_str(int status)
+{
+	char *str = NULL;
+	int size = 0;
+
+	if (status == 0) {
+		str = strdup("DISABLED");
+		return str;
+	}
+
+	for (int i = PLAT_NONE; i <= PLAT_STOPPED; i++) {
+		if (!((1 << i) & status))
+			continue;
+		const char *name = pv_platform_status_string(i);
+		int len = strlen(name);
+
+		char *tmp = realloc(str, len + size + 1);
+		if (!tmp) {
+			free(str);
+			return NULL;
+		}
+		str = tmp;
+
+		memcpy(str + size, name, len);
+		size += len;
+		str[size] = ',';
+		size++;
+	}
+
+	if (str)
+		str[size - 1] = '\0';
+
+	return str;
 }
 
 static off_t _get_size_in_bytes(const char *size_str)
@@ -1197,6 +1283,9 @@ static int _set_config_by_entry(struct pv_config_entry *entry,
 	case WDT_MODE:
 		_set_config_by_entry_wdt_mode(entry, value);
 		break;
+	case PLATFORM_STATUS:
+		_set_config_by_entry_platform_status(entry, value);
+		break;
 	default:
 		pv_log(WARN, "unknown config type %d for key '%s'", entry->type,
 		       entry->key);
@@ -1618,6 +1707,8 @@ static void _format_value_str(char *out, size_t len, config_index_t ci)
 	if (ci >= PV_MAX)
 		return;
 
+	char *status = NULL;
+
 	switch (entries[ci].type) {
 	case BOOL:
 		snprintf(out, len, "%d", entries[ci].value.b);
@@ -1658,6 +1749,17 @@ static void _format_value_str(char *out, size_t len, config_index_t ci)
 		snprintf(out, len, "%s",
 			 _get_wdt_mode_str(entries[ci].value.i));
 		break;
+
+	case PLATFORM_STATUS:
+		status = _get_platform_str(entries[ci].value.i);
+		if (!status) {
+			snprintf(out, len, "couldn't get status");
+			break;
+		}
+
+		snprintf(out, len, "%s", status);
+		free(status);
+		break;
 	default:
 		pv_log(WARN, "unknown type for config entry %s",
 		       entries[ci].key);
@@ -1689,6 +1791,11 @@ static void _add_config_entry_json(config_index_t ci, struct pv_json_ser *js)
 	}
 }
 
+bool pv_config_platform_status_has_alert(int status)
+{
+	return ((1 << status) & entries[PV_PLATFORM_ALERT_STATUS].value.i);
+}
+
 char *pv_config_get_json()
 {
 	struct pv_json_ser js;
@@ -1715,6 +1822,7 @@ static void _print_config_entry(config_index_t ci)
 	const char *key = entries[ci].key;
 	level_t modified = entries[ci].modified;
 	bool secret = entries[ci].secret;
+	char *status = NULL;
 
 	if (secret && modified) {
 		pv_log(INFO, "%s = XXXX (%s)", key,
@@ -1772,6 +1880,16 @@ static void _print_config_entry(config_index_t ci)
 		pv_log(INFO, "%s = '%s' (%s)", key,
 		       _get_wdt_mode_str(entries[ci].value.i),
 		       _get_mod_level_str(modified));
+		break;
+	case PLATFORM_STATUS:
+		status = _get_platform_str(entries[ci].value.i);
+		if (!status) {
+			pv_log(INFO, "could not get %s configuration", key);
+			break;
+		}
+		pv_log(INFO, "%s = '%s' (%s)", key, status,
+		       _get_mod_level_str(modified));
+		free(status);
 		break;
 	default:
 		pv_log(WARN, "unknown type for config entry %s", key);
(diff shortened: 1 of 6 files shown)


asac:
how does the UX experience look like on console? can you post an example for our default stack - aka starter image?

fortizc:
> how does the UX experience look like on console? can you post an example for our default stack - aka starter image?

Sure, this is the start screen with ALL (all events will be reported)
Screenshot From 2026-08-12 11-02-03

asac:
these seem to not have time since boot like dmesg. probably worth adding.

fortizc:
> these seem to not have time since boot like dmesg. probably worth adding.

Something like this?
Screenshot From 2026-08-12 12-26-41

asac:
those lines should include the info that its pantavisor tellig this. i am thinking maybe we can review how console output of services starting on normal/modern linux consoles look like and see if we can somehow copy that approach

fortizc:
> those lines should include the info that its pantavisor tellig this. i am thinking maybe we can review how console output of services starting on normal/modern linux consoles look like and see if we can somehow copy that approach

On that case, we should standardize all messages like the reboot warn, the "shell timeout deferred to xx seconds", etc. Maybe implement something like pv_log() using pv_wall() as "backend"

asac:
yes, lets use pv_log(WALL, ...) or something and then just have a single on/off switch in configs about "put wall on console"

config.c
```diff
@@ -606,6 +611,87 @@ _set_config_by_entry_log_server_outputs(struct pv_config_entry *entry,
...
+
+		if (pv_str_matches(tok, len, "ALL", strlen("ALL"))) {
+			status_code = 0xffff;
+			goto out;
+		}
+
+		int status = pv_platform_status_from_string(tok);
+		if (status == -1) {
+			pv_log(DEBUG,
+			       "Unrecognized plataform status %s changes will "
```
anibalportero, line 644: please rephrase to "... changes will not be reported"

docs/reference/pantavisor-configuration.md
```diff
@@ -97,6 +97,7 @@ This table contains the currently supported list of configuration keys, sorted a
 | `PV_NET_BRDEV` | interface name | `lxcbr0` | set bridge device name |
 | `PV_NET_BRMASK4` | IP mask | `255.255.255.0` | set bridge IPv4 mask |
 | `PV_OEM_NAME` | string | empty | set OEM name for configuration overrides |
+| `PV_PLATFORM_ALERT_STATUS` | comma-separated list of `NONE`, `INSTALLED`, `MOUNTED`, `BLOCKED`, `STARTING`, `STARTED`, `READY`, `RECOVERING`, `STOPPING`, `STOPPED` or `ALL` | `DISABLED` | set container [status](../overview/containers.md#status) transitions that trigger a [console alert](../overview/containers.md#alerts) |
```
anibalportero, line 100: name is a bit misleading for users I'd say. I'd rather use PV_LOG_* (It has something to do with logs even if not with logserver) or PV_ALERT_* maybe.

Also, isn't it a bit overkill to have all these options instead of a single one to disable/enable somehow all of pv_wall?

anibalportero (commented):
not very strong on this, but I would consider if adding groups (group.c:92) and revision (state.c:556) could be useful. 

For example, same way they are measuring plat start times, we could get the full revision READY time (all plats STARTED/READY).

docs/reference/pantavisor-configuration.md
fortizc: I was thinking on this too. At the beginning this is only for container statuses, but we could include other messages like groups, etc. And maybe makes sense (I think so) to have just one on/off switch. My only concern is finishing with a noisy console which makes a bit difficult to read what is happening

config.c
fortizc: ups, typo, I will fix it

fortizc:
@asac @anibalportero I'm looking something to replace the current format and be clear about the origin of the messages:
```
[    7.197800] [PANTAVISOR] [PLATFORM] INFO: platform 'pvr-sdk' status is now INTALLED
```
```
7.197800 ✔ [PV:PLATFORM]: platform 'pvr-sdk' status is now INTALLED
```
```
[    7.197800] PANTAVISOR: PLATFORM: platform 'pvr-sdk' status is now INTALLED
```

docs/overview/storage.md
```diff
@@ -251,6 +251,32 @@ per sink, either as one of the Go time layout constants or as an arbitrary `strf
 string. See [timestamp formats](../reference/logserver-sockets.md#timestamp-formats) for the
 accepted values and the keys that set them.
 
+### Console alerts
```
anibalportero: I don't understand why this feature, which is logs that are directly printed to console, should have a section in storage.md.

docs/overview/containers.md
```diff
@@ -156,6 +156,21 @@ For now, we only support the `ready` signal, which can be used to get to the [RE
 
 Once all containers have met their [status goal](#status-goal), Pantavisor fires the `system-done` [hook point](hooks.md#hook-points), allowing operators to react to the platform reaching its fully-running state.
 
+### Console Alerts
```
anibalportero, line 159: IMO this could have been a single paragraph inside ## Status, this is the only info we would need here:

Every [status](#status) change is also announced on the [device console](../../meta-pantavisor/getting-started/operate/device-access/serial-port.md) as a [console alert](storage.md#console-alerts), one line per transition.

Then the links already tell the rest of the story.

docs/overview/storage.md
```diff
@@ -251,6 +251,32 @@ per sink, either as one of the Go time layout constants or as an arbitrary `strf
...
+
+Some events matter to whoever is sitting in front of a device rather than reading logs after the
```
anibalportero, line 256: "Some events matter to whoever is sitting in front of a device rather than reading logs after the
fact — a container coming up during boot, for instance."

Sorry, I don't understand this phrase.

docs/overview/storage.md
```diff
@@ -251,6 +251,32 @@ per sink, either as one of the Go time layout constants or as an arbitrary `strf
...
+
+Alerts take the short path: straight to the console, bypassing the log server and its
+[output types](#output-types). They are on by default, and can be turned off with
+[`PV_LOG_CONSOLE_ALERTS`](../reference/pantavisor-configuration.md#summary):
+
+```bash
+PV_LOG_CONSOLE_ALERTS=0
+```
+
+Disabling them only silences the console; the same events are still logged normally. See
```
anibalportero, line 276: You mean disabling silences the console but events are still logged in logserver or where?

docs/reference/logserver-sockets.md
```diff
@@ -366,6 +366,57 @@ warning.
 
 See [Output types](../overview/storage.md#output-types) for what each sink is useful for.
 
+## Console alerts
+
+Besides the sinks above, Pantavisor can mirror selected messages straight to the device console.
```
anibalportero, line 371: This is reference, let's keep all these paragraphs shorter. According to docs/CONTRIBUTING.md

| Folder | Answers | Contains | Never contains |
|--------|---------|----------|----------------|
| [`docs/overview/`](overview/index.md) | *How does this work, and why is it built this way?* | Prose, rationale, diagrams, worked examples, algorithm walk-throughs | Exhaustive key / field / endpoint tables — link to reference instead |
| [`docs/reference/`](reference/index.md) | *What exactly are the valid values?* | Tables: every key, every accepted value, every default, every field, every status code | Multi-paragraph explanation — link to overview instead |
| [`docs/tools/`](tools/index.md) | *How do I invoke it?* | CLI synopsis, flags, exit codes, worked invocations | Subsystem theory |
| [`docs/testing/`](testing/index.md) | *How do I test it?* | The pvtest harness, running the suite, authoring tests, the test list | Yocto build instructions — those stay in meta-pantavisor |

asac (changes requested):
Thanks, the single switch and the `PV_LOG_*` naming look right, and the docs are thorough. A few things before this lands:

**1. One call site per alert.** `pv_platform_set_status()` now logs the same string twice, once through `pv_log` and once through `pv_logserver_console_log`. Every future alert source will have to copy-paste its message the same way. That is what `pv_log(WALL, ...)` was meant to avoid: either a level/flag understood by the `pv_log` macro, or a `pv_log_wall()` variant that emits both the log entry and the console line from one call.

**2. Move the helper into `wall.c`.** `pv_logserver_console_log` / `logserver_utils_wall_msg` run in the main process and never touch the logserver; the PR description itself says it was built on `wall.h`. Putting it in `logserver/` drags `buffer.h` into `logserver_utils.c` and makes the API look like it goes through the log server, which the docs then have to explain away. `wall.c` next to `pv_vwall()` is the natural home.

**3. Buffer (suggestion).** The template is well under 100 bytes plus the caller's format. Taking a buffer from the shared log pool for that is more than needed, and on exhaustion the alert silently loses its prefix. A fixed stack buffer is simpler and deterministic. Note also that if `snprintf` ever truncated the composed *format string* it could split a `%` conversion and hand `vdprintf` an invalid format; not reachable with today's 128 KB pool buffers, but with a stack buffer use `SNPRINTF_WTRUNC` and fall back to the plain `pv_vwall(msg, args)` on truncation.

**4. Nits.** `platforms.c` gains `#include <time.h>` and `#include "wall.h"` but uses neither. The PR description refers to `pv_logserver_send_log()` where it means `pv_logserver_console_log()`.

**On the line format:** keep the one you implemented, `[    7.197800] [PANTAVISOR] [platforms] INFO: ...`. It is the only one of the three candidates that carries a level, and it lines up with dmesg.

The `debug.c` `pv_wall` callers (reboot deferral, shell banner) should stay as they are: those are interactive shell UX, not log events, so no timestamp or level there and they stay outside the switch.

wall.h
```diff
@@ -23,9 +23,15 @@
 #ifndef WALL_H
 #define WALL_H
 
+#include <stdarg.h>
+
 void pv_wall(const char *message, ...);
+void pv_vwall(const char *message, va_list args);
```
anibalportero, line 29: this one breaks a bit the pv_wall_* prefix convention we sometimes follow

fortizc:
> Thanks, the single switch and the `PV_LOG_*` naming look right, and the docs are thorough. A few things before this lands:
> 
> **1. One call site per alert.** `pv_platform_set_status()` now logs the same string twice, once through `pv_log` and once through `pv_logserver_console_log`. Every future alert source will have to copy-paste its message the same way. That is what `pv_log(WALL, ...)` was meant to avoid: either a level/flag understood by the `pv_log` macro, or a `pv_log_wall()` variant that emits both the log entry and the console line from one call.
> 
> **2. Move the helper into `wall.c`.** `pv_logserver_console_log` / `logserver_utils_wall_msg` run in the main process and never touch the logserver; the PR description itself says it was built on `wall.h`. Putting it in `logserver/` drags `buffer.h` into `logserver_utils.c` and makes the API look like it goes through the log server, which the docs then have to explain away. `wall.c` next to `pv_vwall()` is the natural home.
> 
> **3. Buffer (suggestion).** The template is well under 100 bytes plus the caller's format. Taking a buffer from the shared log pool for that is more than needed, and on exhaustion the alert silently loses its prefix. A fixed stack buffer is simpler and deterministic. Note also that if `snprintf` ever truncated the composed _format string_ it could split a `%` conversion and hand `vdprintf` an invalid format; not reachable with today's 128 KB pool buffers, but with a stack buffer use `SNPRINTF_WTRUNC` and fall back to the plain `pv_vwall(msg, args)` on truncation.
> 
> **4. Nits.** `platforms.c` gains `#include <time.h>` and `#include "wall.h"` but uses neither. The PR description refers to `pv_logserver_send_log()` where it means `pv_logserver_console_log()`.
> 
> **On the line format:** keep the one you implemented, `[ 7.197800] [PANTAVISOR] [platforms] INFO: ...`. It is the only one of the three candidates that carries a level, and it lines up with dmesg.
> 
> The `debug.c` `pv_wall` callers (reboot deferral, shell banner) should stay as they are: those are interactive shell UX, not log events, so no timestamp or level there and they stay outside the switch.

I guess the changes have been already addressed

PR #794, opened 2026-08-24
Refactor/logserver dgram

fortizc:
**WIP**: Still draft PR 

Replace the connection-oriented model with a connection-less socket. Also, logserver.c was separated into focused units:
    * logserver_socket: create/send/receive helpers.
    * logserver_epoll: helpers for epoll.
    * logserver_fd: manage the fd subscription mechanism.
    * logserver_internal: to share some internal functions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -263,6 +263,10 @@ add_executable(pantavisor
 			logger.h
 			logserver/logserver.c
 			logserver/logserver.h
+			logserver/logserver_epoll.c
+			logserver/logserver_epoll.h
+			logserver/logserver_fd.c
+			logserver/logserver_fd.h
 			logserver/logserver_filetree.c
 			logserver/logserver_filetree.h
 			logserver/logserver_null.c
@@ -285,6 +289,8 @@ add_executable(pantavisor
 			logserver/logserver_rotation.h
 			logserver/logserver_singlefile.c
 			logserver/logserver_singlefile.h
+			logserver/logserver_socket.c
+			logserver/logserver_socket.h
 			logserver/logserver_stdout.c
 			logserver/logserver_stdout.h
 			logserver/logserver_timestamp.c
--- a/logserver/logserver.c
+++ b/logserver/logserver.c
@@ -20,8 +20,11 @@
  * SOFTWARE.
  */
 
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE
+#endif
+
 #include <linux/limits.h>
-#include <sys/epoll.h>
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <sys/socket.h>
@@ -46,15 +49,17 @@
 #include "logserver_singlefile.h"
 #include "logserver_update.h"
 #include "logserver_stdout.h"
+#include "logserver_fd.h"
 #include "logserver_timestamp.h"
 #include "logserver.h"
+#include "logserver_internal.h"
+#include "logserver_epoll.h"
+#include "logserver_socket.h"
 #include "utils/fs.h"
 #include "utils/json.h"
 #include "utils/system.h"
 #include "utils/list.h"
-#include "utils/socket.h"
 #include "utils/pvsignals.h"
-#include "pvctl_utils.h"
 #include "config.h"
 
 #include "pantavisor.h"
@@ -83,21 +88,6 @@ typedef enum {
 	LOG_CMD_TRANSITION,
 } log_cmd_code_t;
 
-struct logserver_fd {
-	char *platform;
-	char *src;
-	int lvl;
-	int fd;
-	struct dl_list list;
-};
-
-struct logserver_conninfo {
-	int fd;
-	pid_t pid;
-	char *cgroup;
-	struct dl_list list;
-};
-
 struct logserver {
 	pid_t pid;
 	pid_t cmd_pid;
@@ -112,11 +102,7 @@ struct logserver {
 	struct logserver_rot rot;
 	// logserver_fd
 	struct dl_list fdlst;
-	// tmp store for fd returned by connect
-	// only if was sent to the fd_sock
-	struct dl_list tmplst;
 	struct dl_list outputs;
-	struct dl_list conninfo;
 };
 
 static struct logserver logserver = {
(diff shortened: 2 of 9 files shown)


asac (commented):
## Review

The split into `logserver_socket` / `logserver_epoll` / `logserver_fd` is a clear improvement — `logserver.c` drops from ~1550 to 1024 lines and the whole accept / conninfo / tmplst machinery disappears. Datagrams are the right model here: the old code was already one-message-per-connection (accept → single read → close), so the stream socket was buying nothing but syscalls. It also fixes syslog on musl/Alpine, where `syslog(3)` only ever tries `SOCK_DGRAM` and previously found a `SOCK_STREAM` `/dev/log`.

Builds clean (one warning, noted below), `clang-format` clean.

Below: one blocker, three majors, then smaller stuff. Also a compatibility note that I think needs to land in the commit message and the docs.

---

### Protocol compatibility

**Payload protocols are all preserved.** Nothing under `logserver/proto/` is touched — `logserver_proto_get()` still auto-detects and dispatches binary LEGACY (`code=0`), binary CMD (`code=256`), RFC 3164, RFC 5424, JSON and key=value with the same detection order and the same parsers.

**The transport change is breaking for `SOCK_STREAM` clients**, though. AF_UNIX requires the socket type to match, so anything doing `socket(AF_UNIX, SOCK_STREAM)` + `connect()` on `/pantavisor/pv-ctrl-log` now gets `ECONNREFUSED`/`EPROTOTYPE`. In-tree there are no such callers left after this PR. Out of tree:

* syslog clients get *better*, not worse — glibc `openlog` tries DGRAM then STREAM, Go `log/syslog` tries `unixgram` then `unix`, Python `SysLogHandler` defaults to `SOCK_DGRAM` with a STREAM fallback. All three used to fall back; now they hit on the first try. musl is DGRAM-only and was simply broken before.
* Hand-written binary / JSON / KV clients that hardcoded `SOCK_STREAM` do break. `docs/reference/logserver-sockets.md` never states the socket type (it already says "datagram" in the JSON/KV/detection sections), so an author had to read pantavisor source to pick STREAM — but some will have.

Suggest a `BREAKING CHANGE:` footer on the commit and an explicit "`SOCK_DGRAM`, one message per datagram" line in `docs/reference/logserver-sockets.md`.

`pv-fd-log` also changed shape, not just type: old was connect-then-send-fd-on-the-connection, new is a single 4-iovec datagram (`plat[50]`, `src[50]`, `int level`, `int add`). That socket isn't bind-mounted into containers, so it stays host-internal.

---

### Blocker — `is_platform` is now silently ignored; container file-log tailers stop delivering

`logserver.c:877` and `logserver.c:941` call `logserver_sck_send_to()`, which hardcodes its destination at `logserver_socket.c:168`:

```c
get_address(LOGCTRL_FNAME, &addr);   /* -> pv_paths_pv_file() -> <rundir>/pv-ctrl-log */
```

The old code picked the path per call:

```c
pvctl_write_to_path(is_platform ? PLATFORM_LOG_CTRL_PATH : path, ...)
```

`PLATFORM_LOG_CTRL_PATH` is `/pantavisor/pv-ctrl-log` (`paths.h:177`) — the bind mount that `plugins/pv_lxc.c:541` sets up **inside the container's mount namespace**. The host rundir does not exist there.

`pvlogger.c:79` passes `is_platform = true`, and `platforms.c:826-841` `setns()`es pvlogger into the container mount namespace for every non-LXC `log_info`. So for any container declaring `logs` entries in run.json (the `/var/log/messages`-style file tailers), every message now `sendto()`s a path that does not exist → `ENOENT` → return value discarded at `logserver.c:877` → logs silently vanish. `pv_logserver_send_vlog` still returns `to_copy`, so pvlogger's own `written <= 0` guard never fires and nothing is reported either.

`logserver_sck_send_to()` needs the destination — a path argument, or plumb `is_platform` through.

### Major — command datagrams can be silently dropped

The client socket is `SOCK_DGRAM | SOCK_NONBLOCK` (`logserver_socket.c:42`) and `pv_logserver_send_cmd` (`logserver.c:941`) discards the return value. Under a log burst the receive queue fills, `sendto` returns `EAGAIN`, and the datagram is gone. The old path was a blocking connected `write()` loop, so this could not happen.

`LOG_CMD_EXIT` is masked by the `pv_system_kill_force` that follows it, but the others are not: a lost `TRANSITION` files subsequent logs under the stale revision, and a lost `START_UPDATE` / `STOP_UPDATE` means `log.tmp` is never renamed (and `pv_logserver_stop_update` then eats its full 5s `pv_fs_path_exist_timeout`). Commands are rare and must not be lossy — send them on a blocking socket, or retry on `EAGAIN` and log the failure.

### Major — `logserver_epoll_wait` can turn a hard error into a 100% CPU spin

`logserver_epoll.c:26-37` branches on `errno` rather than on `ready < 0`, and returns `0` on a real error:

```c
errno = 0;
do {
	ready = epoll_wait(epfd, ev, max, -1);
	if (errno != 0 && errno != EINTR)
		return 0;
} while (ready < 0);
```

`errno` is cleared once, outside the loop, so an `EINTR` followed by a genuine error is handled only by luck. And when a genuine error does surface (`EBADF` after an fd mishap, say), the `0` return makes `logserver_loop()` return immediately, and `while (!(logserver.flags & LOGSERVER_FLAG_STOP))` in `logserver_start_service` spins at full CPU forever. The usual form:

```c
while ((ready = epoll_wait(epfd, ev, max, -1)) < 0 && errno == EINTR)
	;
```

and return `-1` on a real error so the caller can break out of the loop.

### Major — no error reporting anywhere in `logserver_socket.c`

`logserver_internal.h` exists solely to expose `pv_log()` to the new units, and `logserver_socket.c` includes it — but never calls it. Every failure path (`socket`, `bind`, `setsockopt`, `sendmsg`, `sendto`) returns `-1` with no `strerror(errno)`. The old code logged on each of these. Combined with the two findings above, a misconfigured or overloaded logserver now fails completely silently.

---

### Medium

**NULL deref on the `logserver_fd_new` OOM path.** `logserver_fd.c:65,71` `goto err` → `logserver_fd_free()`, which calls `dl_list_del(&lfd->lst)`. But `dl_list_init()` is only reached at line 80, so `lst.next`/`lst.prev` are still NULL from the `calloc`, and `utils/list.h:47` does `item->next->prev = ...` → segfault. Move `dl_list_init(&new_fd->lst)` to immediately after the `calloc`.

**`MSG_TRUNC` / `MSG_CTRUNC` unchecked.** `logserver_socket.c:71` never inspects `msg.msg_flags`. Two consequences specific to datagrams: an oversized log line is now silently cut at `buffer->size` with no indication, and on `MSG_CTRUNC` the kernel has already closed the `SCM_RIGHTS` fd while `logserver_process_fd_req` still treats the datagram as a valid subscribe request.

**Untrusted `msg->len`** (pre-existing, but this is the natural place to fix it). `proto/logserver_binary.c:33` does `end = msg->buf + msg->len` with `len` taken straight off the wire, and any container with the bind-mounted socket can send a 12-byte datagram claiming `len = 0x7fffffff`; the `strnlen`/`memccpy` walk then runs off the buffer. Equally true on the stream socket, so not a regression — but `logserver_get_log_data` now knows the exact datagram length, so clamping `msg->len` to `received - sizeof(struct logserver_msg)` before parsing is a two-line fix that wasn't as easy before.

### Minor

* `logserver_socket.c:36-44` — the sender socket is unconnected, so it can be created once and cached rather than `socket()` + `sendto()` + `close()` on every single log line (3 syscalls where 1 would do, on the hottest path in the process).
* `logserver_socket.c:130` — `ssize_t ret = sendmsg(sock, &msg, 0) > 0;` assigns a boolean to something named `ret`, then `return ret ? 0 : -1`. Also drops `errno` before the caller could use it. Prefer an explicit `ssize_t n = sendmsg(...)`.
* `logserver_socket.c:39` — unused `struct sockaddr_un addr` in `client_socket()`; `gcc -Wall` warns on it. The `errno = 0;` just above is dead too.
* `logserver.c:479` — the add path doesn't check `logserver_fd_fetch_by_data()` first, so a re-subscribe for the same plat/src leaves the previous fd in `fdlst` and in the epoll set indefinitely.
* `logserver_fd.c:150-175` — the 3-attempt `EAGAIN` retry busy-loops with a fresh `pv_buffer_get()` per attempt. With level-triggered epoll, returning on the first `EAGAIN` is both simpler and correct.
* `logserver.c:747` — `logserver_rfc_create_socket()`'s return is discarded; it's a bare `symlink()`, so a second run fails `EEXIST` with no diagnostic. The old code warned here.
* `CMakeLists.txt` — `logserver/logserver_internal.h` is missing while every other header in the list is present.
* `pvctl_utils.c` — `open_socket()` / `pvctl_write()` / `pvctl_write_to_path()` now have zero callers and still speak `SOCK_STREAM`. Either delete them, or repoint them at the new dgram helper — which is also where the blocker above gets fixed.

fortizc:
I guess I made a mistake not setting this PR as a draft PR. While most of the comments in the review are true some of them are because the incomplete state of the PR and first one the "blocker" isn't really true (those functions are always called only form Pantavisor) 
As a fix, I added the "WIP" in the PR description, once the PR will be ready for review I remove the "tag"

PR #810, opened 2026-09-07
feat(update): honor a Hub-side cancel while queued or downloading

asac:
## Summary
- An owner could cancel a step on the Hub, but the device never looked at its own step again once it had picked it up, and its next progress PUT overwrote the `CANCEL`. A long download could not be stopped from the cloud at all.
- While an update is `QUEUED` or `DOWNLOADING`, the device now polls its own step on every download tick (~6 s). On `CANCEL` it aborts the in-flight object and metadata requests, keeps the partial objects on disk (content-addressed, so a later re-post resumes them), reports `CANCEL` with `Cancelled as requested by owner` and returns to idle on the previous revision.
- The download progress PUT of a tick is only sent from the step-poll callback, after the poll came back without a cancel. Sending both in the same event-loop pass (as the first version did) meant the PUT always overwrote a cancel that landed since the previous tick, so the device could never see it. A cancel landing in the short window between the poll and the PUT is still overwritten and has to be issued again; progress PUTs stay unconditional on the Hub on purpose so a non-adhering device stays visible.
- `CANCEL` is a final status but not a failed one: the bootloader was never touched. After `INPROGRESS` the device no longer looks at its step; a Hub-side cancel or wontgo there only stops a device without local knowledge of the step from re-fetching and retrying it. Documented in `docs/overview/updates.md`.
- Carries the test-only `PH_UPDATER_DOWNLOAD_RATE_LIMIT` knob so the cancel window can be widened deterministically in pvtest. It paces reads in pantavisor itself (pause `EV_READ` after each chunk, re-enable from a one-shot timer), so it works with any TLS backend and needs no libevent patch; at the default `0` no pacer is allocated and no callback is registered.

## Prerequisite for
- #770 (object-download-resume) rebases on top of this once merged and drops its own token-bucket knob and the meta-pantavisor libevent mbedtls patch. Rationale: no super-long download should be possible before it can be properly cancelled.

## Companions
- pantahub-base !415 (merged, deployed on api.pantahub.com): accept owner cancel in `QUEUED`/`DOWNLOADING`/`INPROGRESS` (was `NEW` only) and with session tokens; `wontgo` stays supported for existing devices, marked deprecated. Device progress PUTs stay unconditional on purpose so a non-adhering device is visible.
- meta-pantavisor #486: `remote/lifecycle/download-cancel-mid-transfer` + testplan (re-issues a cancel lost to the poll/PUT window, retries on stderr only).

## Test plan
- [x] appengine build via kas-container (docker-x86_64-scarthgap + workspace), clean sstate
- [x] `local/lifecycle/seq-non-reboot-updates`
- [x] `remote/lifecycle/update-retries-pv-crash` (download state with the new polling, real Hub)
- [x] `remote/lifecycle/download-cancel-mid-transfer` end to end against production Hub with !415 deployed: 100 MB object throttled to 1 MB/s, partial on disk before the cancel, device polled its step 3 times, reported `CANCEL`, stayed on the previous revision, partial kept, Hub status stays `CANCEL`

## Results (docker-x86_64 appengine, real Hub, 2026-09-08, rebased on master)

| Test | Result |
|---|---|
| `local/lifecycle/seq-non-reboot-updates` | PASSED (94 s) |
| `remote/lifecycle/update-retries-pv-crash` | PASSED (204 s) |
| `remote/lifecycle/download-cancel-mid-transfer` | PASSED (188 s), second run (first run RECORDED the golden byte-identical to the hand-written one, cancel honored on the first attempt) |

Before the poll-then-PUT ordering fix the same end-to-end run timed out: the Hub answered the cancel with 200, the next tick's progress PUT overwrote it and the device read back `DOWNLOADING` every 6 s.


docs/reference/pantavisor-configuration.md
```diff
@@ -68,6 +68,7 @@ This table contains the currently supported list of configuration keys, sorted a
 | `PH_METADATA_DEVMETA_THRESHOLD` | percent | `1` | relative change a numeric [device metadata](pantavisor-metadata.md#change-thresholds) field needs before it triggers a push of its own |
 | `PH_METADATA_USRMETA_INTERVAL` | time (in seconds) | `5` | set refresh interval for [user metadata](../overview/storage.md#user-metadata) from [Pantacor Hub](../overview/remote-control.md#pantacor-hub) |
 | `PH_ONLINE_REQUEST_THRESHOLD` | number of failures | `0` | number of failed requests to [Pantacor Hub](../overview/remote-control.md#pantacor-hub) allowed to still consider device online |
+| `PH_UPDATER_DOWNLOAD_RATE_LIMIT` | bytes per second | `0` (unlimited) | throttle object-download throughput; test/debug knob for reliably exercising resume/interruption behavior, not a production traffic-shaping feature |
```
anibalportero, line 71: link in description to updates.md#downloading would be nice

PR #818, opened 2026-09-11
test(pvtest): add real device support to volatile model

anibalportero:
This PR adds real device support to `--model volatile` from test.docker.sh and test.native.sh.

To make executing each test in a freshly flashed device, we have added the `flash=` key to `devices.txt` manifest, mandatory in volatile model against real device. This is a hook  that can be used to set a command that will be executed during the initialization of the test. This will allow the user to run a script that flashes, installs initial containers and sets boot config env following the specifics of each board and lab setup.


ferlzc:
@anibalportero are you still working on this feature? Let me know when you finished so I can tested here on my lab since I can flash my devices.

